PR #5, opened 2015-03-11
Support for early mutators

colincross:
This patch set is an alternative to the namespaced modules we discussed previously, as well as greatly simplifying dependencies that vary based on a mutated axis.  It:
- turns module groups into a list of modules, moving all other data into the individual modules
- iterates through modules directly instead of through groups and then modules, but still maintains the module ordering within a group
- adds EarlyMutators that can split modules before any dependencies have been resolved
- adds AddVariantDependency that allows a DynamicDependerModule to specify a variant it wants to depend on.


colincross:
FYI, this is on top of the bugfixes branch, but github seems to order them by date instead of commit dependencies in the list above so they are hard to pick out.  They might be easier to review after the bugfixes branch is merged, and if not I'll update the pull request once the bugfixes branch is merged.

PR #12, opened 2015-03-21
Update import paths to include github

jgennis:


colincross:
LGTM, soong still compiles after fixing the import paths

PR #23, opened 2015-04-24
pathtools.Glob bugfixes and features

colincross:
This patch set fixes a bug in pathtools.Glob that caused it to miss some directory dependencies, and adds support for recursive globs using "**", and GlobWithExcludes.


colincross:
Found a bug, closing for now

PR #32, opened 2015-06-11
Add bpfind tool

danw:
Blueprint files aren't necessarily called 'Blueprints' anymore, with the use of subname= and build=. Add a tool that can print a list of parsed blueprints.


danw:
closing this for now - there's not a pressing need for it, and the implementation would be different with the new changes that have gone into blueprint.

PR #33, opened 2015-06-17
Allow overriding top level Blueprints name in bootstrap.bash

colincross:
Allow the project bootstrap.bash to override the top level Blueprints
file name in case if it is not ${SRCDIR}/Blueprints.

Change-Id: Id5acba5f6fd4ad64db1fde41eae1955f8a1cd2d5


danw:
I made a similar change in #31 and https://android-review.googlesource.com/#/c/154255/

PR #34, opened 2015-06-18
Expose the actual Blueprints file to the module ctx

danw:


danw:
I can use SingletonContext.BlueprintFile instead

PR #41, opened 2015-07-07
Fix a sub-Blueprints deps bug.

jgennis:
This change fixes a bug where directories that could contain a Blueprints file
but currently do not were not being added as a dependency.  This meant that if
a Blueprints file were added to that directory it would not automatically get
picked up by the build.


colincross:
LGTM

PR #43, opened 2015-07-09
Add option to run bootstrap.Main, but not write the ninja file

danw:
So that everything is evaluated, but we don't actually write out the
ninja file. This is intended for use by tools to inspect modules that
have been fully created.


danw:
Replaced with #77

PR #44, opened 2015-07-11
Fix bugs related to local vs. inherited variables

colincross:
The Go race detector found a race condition in the parser, which
 highlighted a few related bugs. A variable could be defined but
not referenced in a Blueprints file, then appended to in multiple
subdirs= Blueprints files.  The race detector caught the multiple
writes to assignment.Referenced from the parsers for the subdirs
Blueprints files, but multiple appends would be much more serious.

To fix this, keep local and inherited variables separate in the
Scope object and export that info to the parser.  Disallow
appending to non-local variables, which was already the intended
behavior.  Only update the referenced boolean for local variables.
Together, this should prevent all writes to Assignment objects
from parsers other than the one that created them.

Also improves the error handling code and some error messages.

Change-Id: Idb4f7d2e61bbe28d90b93074764e64e60d1eba8f

### DIFF
--- a/context.go
+++ b/context.go
@@ -465,12 +465,12 @@ func (c *Context) parse(rootDir, filename string, r io.Reader,
 	}
 	file.Name = relBlueprintsFile
 
-	subdirs, subdirsPos, err := getStringListFromScope(scope, "subdirs")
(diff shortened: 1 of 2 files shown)


danw:
Other than the style nit, LGTM

PR #46, opened 2015-07-23
Multi-stage bootstrapping

danw:
The details are in the last two commit messages - short summary is that this splits the current bootstrap build stage in two, so that we can check in the build.ninja.in for only minibp. That lets us dynamically generate the primary builder based on what's in the tree (which will be a future change).


colincross:
The travis failure is due to the ninja 1.6.0 requirement, should pass after #47 is merged

PR #47, opened 2015-07-24
Updates to travis config -- faster; ninja v1.6.0

danw:
* Fix running travis on forks by moving the code around to always be gopath/src/github.com/google/blueprint
* Migrate to the container based travis setup. This should make our builds start faster according to the docs, since they can scale containers faster than VMs.
* Use ninja v1.6.0, built from source (and cached between builds).


colincross:
LGTM, Jamie?

PR #48, opened 2015-08-01
Support go 1.5

danw:
The go compiler and linker changed in v1.5 -- to 'go tool compile' and
'go tool link' instead of 6g and 6l. Move the selection logic to
bootstrap.bash, and have it use compile/link if 6g/6l are missing. This
way the build.ninja.in will continue working with either go 1.4 or 1.5.

Travis and the test suite will fail under 1.5, since the build.ninja.in
is still generated with 1.4, and the function names in the comments
differ between 1.4 and 1.5.

### DIFF
--- a/bootstrap.bash
+++ b/bootstrap.bash
@@ -59,6 +59,21 @@ EXTRA_ARGS=""
 # If RUN_TESTS is set, behave like -t was passed in as an option.
 [ ! -z "$RUN_TESTS" ] && EXTRA_ARGS="$EXTRA_ARGS -t"
 
+GOTOOLDIR="$GOROOT/pkg/tool/${GOOS}_$GOARCH"
+GOCOMPILE="$GOTOOLDIR/${GOCHAR}g"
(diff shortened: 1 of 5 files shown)


colincross:
Also see https://github.com/colincross/blueprint/commit/beb18f4bba3b0ee2094dd93f12ba12ccbfc64c3c which does the same thing but without backwards compatibility.  The primary difference is that I used "go tool compile" instead of executing the compile binary directly, which might be better for forwards compatibility.

danw:
Yeah, I looked at that first, but wanted backwards compatibility. 

6g/6l was also available via "go tool", at least in 1.4. We just need to pick a method. It's a simple, but not trivial change, since we need to make sure the dependencies are proper.

colincross:
Might as well keep that as a separate change then

danw:
Rebased on master to fix build.ninja.in conflicts. Also renamed the "gc" rule to "compile"

PR #49, opened 2015-08-01
Implement go bootstrap plugins

danw:
Now that we have multi-stage bootstrapping, we can make the primary
builder build more dynamic. Add the concept of plugins that will be
linked and loaded into bootstrap_go_binary or bootstrap_go_package
modules. It's expected that the plugin's init() functions will do
whatever registration is necessary.

Example Blueprint definition:

    bootstrap_go_binary {
      name: "builder",
      ...
    }

    bootstrap_go_package {
      name: "plugin1",
      pluginFor: ["builder"],
    }

A package may specify more than one plugin if it will be inserted into
more than one go module.

### DIFF
--- a/Blueprints
+++ b/Blueprints
@@ -128,3 +128,8 @@ bootstrap_core_go_binary(
     name = "choosestage",
     srcs = ["choosestage/choosestage.go"],
 )
+
+bootstrap_go_binary{
+    name = "loadplugins",
+    srcs = ["loadplugins/loadplugins.go"],
+}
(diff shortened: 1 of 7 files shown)


danw:
It may be useful to support plugins for bootstrap_go_packages as well, so that a single package can collect a number of plugins, then several binaries could use that package without needing to be added to every plugin's list.

danw:
And now inserting plugins into bootstrap_go_packages works as well. It also verifies that plugins are only linked into non-core modules (since otherwise the bootstrap ninja file would change)

PR #51, opened 2015-08-13
bpdoc: Cache property struct docs by full package path

colincross:
The full package path is no longer stored in the name variable, pass the
pkg variable into getDocs and putDocs so that types with the same name
in different packages don't collide.

Change-Id: I7ecd98815487467dbaadc8a318fa43b123fc83c1


danw:
LGTM

PR #53, opened 2015-08-26
bootstrap: Added phony rule for bootstrapManifest.

yuchenericwu2:
This is for stageMain and prevents build.ninja.in from being
accidentally deleted.

Change-Id: I9675874c61c3db0bf34c47498968623b8c183e95


danw:
LGTM

PR #54, opened 2015-08-26
context: Changed singletons to execute in order.

yuchenericwu2:
Replaced map with list so that singletons will execute in the order
that they were listed.

Change-Id: I62919b14110b2b6f1f31c18f249ca64b85898ab1


colincross:
LGTM

PR #58, opened 2015-09-18
"Stop using bootstrap.bash -b for separate builddir

danw:
It's difficult for wrapping scripts to handle -b properly. Just pass
BUILDDIR instead, which is easier to handle. This still accepts -b, so
that incremental builds work across this change.


danw:
This allows us to make this change: https://android-review.googlesource.com/171821 and stop parsing arguments in our bootstrap.bash wrapper

PR #62, opened 2015-10-29
Add helpers for extending properties to proptools

colincross:
Add helper functions to append or prepend properties in property
structs.


colincross:
I'm going to replace this with a larger pull request

PR #63, opened 2015-10-29
Deprecate EarlyMutator and DynamicDependencies

colincross:
DynamicDependencies can be implemented more flexibly by a
BottomUpMutator.  If there are no DynamicDependencies, then
EarlyMutators are identical to BottomUpMutators.  Deperecate both, and
reimplement DynamicDependencies inside a BottomUpMutator that is
guaranteed to be registered first.


danw:
LGTM

PR #64, opened 2015-11-01
Generic append properties, and *bool support

colincross:
It is common for build logic to append the values of two property structs together.  The first patch provides generic AppendProperties and PrependProperties to append property structs of the same type, and AppendMatchingProperties and PrependMatchingProperties to append the values in a property struct to any matching properties in a list of destination property structs.

The second and third patches add tests and fix bugs in CloneProperties, CloneEmptyProperties, and ZeroProperties.

To make AppendProperties easy to use appending an unset property should not change the destination property.  This means appending a zero value property must be a no-op.  The only way to accomplish this for a bool property is to OR the two values together, but this is usually not the semantics needed by the build logic.

The fourth patch fixes lists to properly differentiate empty and unset.

The fifth patch adds support for *bool properties with identical syntax to bool properties, but with replace-if-not-nil append semantics.  *bool also provides a better replacement for ContainsProperty, which can only tell if the value was set by a blueprints file, and not set later by a mutator.  *string support is added for the same reason, but has the same append semantics as string.


danw:
LGTM

colincross:
I added one more patch to recursively compare types in property structs, which is useful when using AppendProperties on lists of structs of unknown types.

danw:
LGTM

PR #65, opened 2015-11-03
Improve PropertyErrorf

colincross:
PropertyErrorf gets called from build logic, and shouldn't panic if the
build logic reports an error on an unset property.  Fall back to using
the module position if the property wasn't set.

Also put the property name into the error message.


danw:
LGTM

PR #66, opened 2015-11-03
Add OtherModuleName and OtherModuleErrorf to mutators

colincross:
TopDownMutators can visit dependencies, provide OtherModuleName and
OtherModuleErrorf so they can query and report errors on those
dependencies.


danw:
LGTM

PR #67, opened 2015-11-04
Sort reverse dependencies

colincross:
AddReverseDependencies would add modules the target's dependency list in
a non-deterministic order based on the order the modules were parsed.
Redefine AddReverseDependencies to collect dependencies until the end of
the mutator and then add them sorted by name.


danw:
LGTM

PR #68, opened 2015-11-04
Print useful error for self-dependency

colincross:
beforeInModuleList panic'd when checking if a variant was before itself
in the module list.  Return a real error instead of calling
beforeInModuleList, and also fix beforeInModuleList to return false for
the same module.


danw:
LGTM

PR #69, opened 2015-11-04
Don't rely on deps property position

colincross:
Dependency errors were prefixed with ??? because they were associated
with the position of the "deps" property, which is often not used.  Use
the position of the module instead.


danw:
It would be nice to have the actual position, but I guess that would require an API change. LGTM

PR #70, opened 2015-11-08
Add blank line after build rules, for human readability.

dje4g:


colincross:
Any patch that changes the output of the bootstrap stage needs to regenerate the build.ninja.in file.  To do that, bootstrap a vanilla Blueprint build and then regenerate the build.ninja file.  From the blueprint directory:
```bash
mkdir /tmp/blueprint
cd /tmp/blueprint
$OLDPWD/bootstrap.bash
ninja
$OLDPWD/bootstrap.bash -r
```
This should update build.ninja.in, add that to your patch and re-push it.

PR #74, opened 2015-11-09
Add blank line after build rules, for human readability.

dje4g:
bash$ sh tests/test.sh
passes now. Thanks.


colincross:
Unfortunately this one is going to fail too, because Travis uses go 1.4, which names the inline functions differently.  Let me see about switching it to 1.5.

Also, you can force push your new patches over the old branch and github will update the old pull request.

PR #76, opened 2015-11-18
Add variant methods to SingletonContext and Context

colincross:
Add VisitAllModuleVariants, PrimaryModule, and FinalModule to
SingletonContext so singletons and tools can deal with modules with
multiple variants.


danw:
LGTM

PR #77, opened 2015-11-18
Add option to run bootstrap.Main, but not write the ninja file

colincross:
So that everything is evaluated, but we don't actually write out the
ninja file. This is intended for use by tools to inspect modules that
have been fully created.

Change-Id: I7d1ac436a28ada7d4c1eb2311963aeb1f9df3d4c

### DIFF
--- a/bootstrap/command.go
+++ b/bootstrap/command.go
@@ -42,7 +42,7 @@ var (
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


colincross:
Is passing -d on the command line to write a depFile not sufficient?

danw:
No, because that happens in bootstrap.Main after your new return statement.

PR #78, opened 2015-11-18
Add common implicit deps to Rules

danw:
For implicit dependencies that will be common to all users of a Rule,
add a new field 'CommandDeps' to the RuleParam. This is a list of
strings to be prepended to the implicit dependencies in each BuildParam.

This lets us have the dependencies declared next to where they are used,
instead of duplicated in areas that may be far apart.

I looked at passing this information down to ninja too, but it only
saves us a few percent of ninja file, and requires a modification to the
ninja file format.


danw:
Hmm, I'm actually seeing some failures with the latest patch. I'll reopen once I've fixed them.

danw:
Fixed (variable scoping was wrong in live_tracker.go)

colincross:
LGTM

PR #79, opened 2015-11-18
End removing directories on ENOTDIR

colincross:
An abandoned directory may have been replaced with a file, give up on
removing the directory on ENOTDIR.


danw:
LGTM

PR #80, opened 2015-11-18
Make removing abandoned files optional

colincross:
Build logic can now implement a RemoveAbandonedFiles, and the bootstrap
logic will only remove abandoned files if that method returns true.
Leaving the method unimplemented will result in the existing behavior of
always removing abandoned files.


danw:
LGTM

PR #81, opened 2015-11-18
Allow multiple calls to ctx.SetBuildDir

colincross:
ctx.SetBuildDir is called by the bootstrap singleton, which is always
last.  In order to allow the primary builder to set the directory for
.ninja_log/.ninja_deps, remove the panic on multiple calls to
ctx.SetBuildDir, and always used the value passed by the first caller.


danw:
LGTM

PR #82, opened 2015-11-19
Use context builddir for removing abandoned files

colincross:
Removing abandoned files needs to know where the .ninja_log file is
stored.  Export the ninja builddir value from Context and use it to
determine the .ninja_log path in any stage.

The ninja builddir (where ninja stores its .ninja_log and .ninja_deps
files) and the bootstrap.BuildDir (where build output files are written)
are distinct, so to reduce confusion replace SetBuildDir with
SetNinjaBuildDir.


danw:
LGTM

PR #83, opened 2015-11-21
Support embedded anonymous property structs

colincross:
Allow property structs to contain anonymous embedded structs and
interfaces.  Properties in an anonymous embedded struct or interface are
treated as if they were properties in the embedding struct.


danw:
LGTM

colincross:
Added one more patch to prevent errors when the docs generator came across anonymous embedded structs.

danw:
LGTM

PR #84, opened 2015-11-30
Allow wrapping of PackageContext

danw:
Turn PackageContext into an interface so that build systems can wrap it
to add more custom helpers.

This does introduce an API change, though it should be fairly simple.
NewPackageContext used to provide an opaque *PackageContext struct, now it
provides a PackageContext interface.

### DIFF
--- a/context.go
+++ b/context.go
@@ -82,7 +82,7 @@ type Context struct {
 	ignoreUnknownModuleTypes bool
 
 	// set during PrepareBuildActions
-	pkgNames        map[*PackageContext]string
+	pkgNames        map[*packageContext]string
 	globalVariables map[Variable]*ninjaString
 	globalPools     map[Pool]*poolDef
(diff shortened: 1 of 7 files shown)


danw:
Because the ModuleContext.(Variable|Rule|Build) functions take a *PackageContext directly, so we'd need to unwrap the struct on every one of those calls.

colincross:
True, but aren't you already going to wrap the ModuleContext functions?

danw:
We cannot change the function signatures, since we embed blueprint.ModuleContext in the AndroidModuleContext interface. So we can wrap them in different names, but the originals still must exist in our interface.

They're also on SingletonContext, which we don't wrap currently, but we don't use many of those, so unwrapping wouldn't be too bad.

colincross:
LGTM

PR #85, opened 2015-12-10
Add SingletonContext.Failed()

danw:
Identical to ModuleContext.Failed(), just returns whether there has been
an error added to the context.


colincross:
LGTM

PR #86, opened 2015-12-11
Add build wrapper to do stage selection before ninja

danw:
This wrapper script can be used instead of ninja to ensure the build
won't get stuck building the primary builder.

An example of when this would happen:

1. Do a successful build
2. Sync/make a change in the primary builder (soong, etc) that depends
   on a blueprint change.
3. The next build would notice that change, and rewind to the primary
   stage to rebuild the builder. That build would fail.
4. Sync/fix the blueprint code.
5. The next build would still fail. The bootstrap stage would need to
   be run in order to fix the primary stage, but we're still stuck in the
   primary stage. The only way to switch stages is to successfully
   complete everything required to choose the next stage.

This generally isn't a problem in the main stage, since there is no code
being built in the dependency chain leading up to stage selection.

Any existing wrappers (like soong) can execute this wrapper (optionally
skipping ninja execution) if they're worried about the above situation.
This isn't strictly required -- running ninja directly will still work
in most cases, you'll just need to re-run bootstrap.bash if you get into
a bad state.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ before_script:
 script:
     - go test ./...
     - mkdir stage
-    - cd stage 
+    - cd stage
     - ../bootstrap.bash
-    - ninja
+    - ./blueprint.bash
     - diff -us ../build.ninja.in .bootstrap/bootstrap.ninja.in
     - ../tests/test.sh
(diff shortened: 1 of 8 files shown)


colincross:
LGTM

PR #87, opened 2015-12-17
Add optional_subdirs variable support

colincross:
When building a subset of a source tree, some of the subdirectories
mentioned in the subdirs variable may be missing.  Add support for an
optional_subdirs variable, which won't error out if the requested
directory is missing.


danw:
LGTM

PR #88, opened 2015-12-18
Allow primary builder to handle missing dependencies

colincross:
In some cases the primary builder may need to handle missing
dependencies.  Add Context.SetAllowMissingDependencies to cause
Blueprint to store the list of missing dependencies without immediately
emitting an error, and ModuleContext.GetMissingDependencies to return
the missing dependencies to the primary builder.  If the primary builder
does not call ModuleContext.GetMissingDependencies Blueprint will emit
dependency errors.


danw:
LGTM

PR #89, opened 2015-12-18
Add ModuleSubDir to Context and SingletonContext

colincross:
Singletons may need ModuleSubDir, for example to implement a stable sort
on modules.  Add ModuleSubDir to SingletonContext and Context.


danw:
LGTM

PR #90, opened 2015-12-18
Add AddNinjaFileDeps to PackageContext

danw:
Soong needs to add dependencies during a VariableFunc. (we're doing a glob)


colincross:
LGTM

PR #91, opened 2016-01-05
AppendProperties: Replace *strings instead of appending

danw:
This makes *string values act like *bool values -- instead of appending
or prepending the contents of the string, the entire string is replaced.
The use case here is for overriding filenames, where appending doesn't
work.


colincross:
LGTM

PR #93, opened 2016-01-13
docs: Allow propery structs to contain non-struct ptrs

danw:
Other parts of blueprint have started allowing pointers to strings and
booleans. This code silently allowed them because of the nil check, but
it doesn't work if the module factory sets a default value.


colincross:
LGTM, it is no worse than what is there but we should fix the type eventually

danw:
Passed travisci on my fork: https://travis-ci.org/danw/blueprint/builds/102211094

PR #94, opened 2016-03-03
walkDeps: Fixed bug on non-walked, visited deps.

yuchenericwu2:
visited map was not being updated when the dependency was being
visited, only if it was being walked. This led to cases where
duplicate traversals were possible.


yuchenericwu2:
It's necessary because logic for generating ninja rules typically exists within the visit() function. walk() is only used for continuing to traverse that node's dependencies.

There is a bug in the existing code that if a node is visited but not walked, it won't be marked as visited. Therefore duplicate ninja rules can be generated for that node if it is visited again.

PR #95, opened 2016-03-17
walkDeps: Fixed bug with the visit nodes.

yuchenericwu2:
The last fix actually introduced a new bug. The problem was that it
changed the meaning of visit()'s return value from whether the node
was visited to whether the node's dependencies should be walked. In
actuality, we need a way to express both of these meanings.

This breaks the API for walkDeps, but was anyone else actually
using this function?

### DIFF
--- a/context.go
+++ b/context.go
@@ -1926,7 +1926,7 @@ func (c *Context) processLocalBuildActions(out, in *localBuildActions,
 }
 
 func (c *Context) walkDeps(topModule *moduleInfo,
-	visit func(Module, Module) bool) {
+	visit func(Module, Module) (bool, bool)) {
 
 	visited := make(map[*moduleInfo]bool)
 	var visiting *moduleInfo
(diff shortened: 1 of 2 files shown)


colincross:
This API is getting really messy - trying to document the meaning of the two bools returned by the visit function in a way that is easily understood seems impossible.  You can probably solve your problem using one or more TopDownMutators instead, and keep all of the custom logic about visiting and walking deps in your build logic.

yuchenericwu2:
Is it possible to obtain edge information (i.e. both parent and child node)
while traversing with a TopDownMutator? That was the reason I originally
needed to add walkDeps.

On Thu, Mar 17, 2016 at 10:23 AM colincross <notifications@github.com>
wrote:

> This API is getting really messy - trying to document the meaning of the
> two bools returned by the visit function in a way that is easily understood
> seems impossible. You can probably solve your problem using one or more
> TopDownMutators instead, and keep all of the custom logic about visiting
> and walking deps in your build logic.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/google/blueprint/pull/95#issuecomment-197985709>
>
-- 
Yuchen Wu
Software Engineer, Help
646-671-3992
solveforx.com

colincross:
Yes - TopDownMutator calls your mutator function for each module, and then the mutator function can call VisitDirectDeps to call a function on each child.  That function can be a closure that has access to both the module passed to the mutator function and the module passed to the visit function.

PR #96, opened 2016-03-24
Fix package names for executables.

TKilbourn:
Go 1.6 fixed a bug where compiling with the -pack flag allowed non-main
packages to link.
https://github.com/golang/go/issues/13468

This change uses package main for choosestage and gotestmain.


danw:
Thanks!

PR #97, opened 2016-03-29
Add SingletonContext.Eval to evaluate ninja strings

danw:
Our use case is to write out some configuration variables to give to our
legacy build system. But in order to do that, we either need to keep a
Go copy of all of the configuration, and still have all the calls to
convert them to ninja variables, or evaluate our ninja variables.


colincross:
LGTM

PR #98, opened 2016-04-12
Add dependency tags, clone all modules after mutators, support AddInterVariantDependency

colincross:


context.go
```diff
@@ -1312,6 +1312,35 @@ func (c *Context) addVariationDependency(module *moduleInfo, variations []Variat
...
+
+	var fromInfo, toInfo *moduleInfo
+	for _, m := range origModule.splitModules {
+		if m.logicModule == from {
+			fromInfo = m
+		}
+		if m.logicModule == to {
+			toInfo = m
+			if fromInfo != nil {
+				return []error{&Error{
```
danw: Why an error here, and a panic below? Both are go logic errors, not user blueprint error (so the position isn't helpful)

danw:
Commit message typo on the middle commit: retreivew

module_ctx.go
```diff
@@ -271,7 +271,7 @@ func (m *baseModuleContext) OtherModuleDependencyTag(logicModule Module) Depende
 func (m *baseModuleContext) VisitDirectDeps(visit func(Module)) {
 	defer func() {
 		if r := recover(); r != nil {
-			panic(newPanicErrorf(r, "VisitDirectDeps(%s, %s, %s) for dependency %s",
+			panic(newPanicErrorf(r, "VisitDirectDeps(%s, %s) for dependency %s",
```
danw: Shouldn't these two be moved to the previous commit? (Though it doesn't really matter)

danw:
LGTM

PR #99, opened 2016-05-05
Support ExtendProperties that can append or prepend

colincross:
ExtendProperties is the same as AppendProperties or PrependProperties,
but takes a function that determines whether each property should be
appended or prepended.

Change-Id: I26e400d56d75a88bab9c27c382ee5321bc623ee5


danw:
LGTM

colincross:
Added tests

danw:
LGTM

PR #100, opened 2016-05-26
Add blueprint_go_binary for user-run tools

danw:
Move these tools from $buildDir/.bootstrap/bin to $buildDir/bin
(configurable by the primary builder).

Also delay building them until the main stage, and give them a phony
target "blueprint_tools".


colincross:
LGTM

PR #101, opened 2016-06-02
bpdoc improvements

colincross:
Improve readability in bpdoc by removing naming stutter.

Fixes the documented type for properties stored in pointers, previously printed as "Type: *ast.StarExpr"


danw:
LGTM

PR #102, opened 2016-06-02
Allow tests to mock filesystem

colincross:
Tests can call Context.MockFileSystem to pass in a map of filenames to
contents.

Change-Id: Idd67c68f7cb43bc2117cc78336347db7e2f21991


danw:
LGTM

PR #103, opened 2016-06-08
Refactor blueprint parser nodes to an interface

colincross:
Refactor the blueprint parser Value object, which contained a Type enum
and members to hold every possible type, into an interface (now called
Expression).  Rename the existing Expression object that represented a binary
operator Operator.

Also adds and fixes some new printer test cases with mulitline expressions.

Change-Id: Icf4a20f92c8c2a27f18df8ca515a9d7f282ff133

### DIFF
--- a/parser/modify.go
+++ b/parser/modify.go
@@ -14,47 +14,38 @@
 
 package parser
 
-func AddStringToList(value *Value, s string) (modified bool) {
-	if value.Type != List {
-		panic("expected list value, got " + value.Type.String())
-	}
+import "fmt"
 
-	for _, v := range value.ListValue {
-		if v.Type != String {
-			panic("expected string in list, got " + value.Type.String())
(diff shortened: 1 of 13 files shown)


parser/modify.go
```diff
@@ -14,47 +14,50 @@
 
 package parser
 
-func AddStringToList(value *Value, s string) (modified bool) {
-	if value.Type != List {
-		panic("expected list value, got " + value.Type.String())
+import "fmt"
+
+func AddStringToList(value Expression, s string) (modified bool) {
```
danw: Why not take a List here instead of an Expression? (and in RemoveStringFromList)

danw:
Other than the small nit (which could be a later change), LGTM

colincross:
Fixed

danw:
LGTM

PR #104, opened 2016-06-13
More parser cleanups

colincross:
Make a few more refactoring changes to blueprint/parser to prepare for utilities to modify an AST.

### DIFF
--- a/parser/ast.go
+++ b/parser/ast.go
@@ -20,8 +20,16 @@ import (
 	"text/scanner"
 )
 
+type Node interface {
+	// Pos returns the position of the first token in the Expression
+	Pos() scanner.Position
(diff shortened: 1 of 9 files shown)


parser/ast.go
```diff
@@ -33,13 +33,13 @@ type Assignment struct {
 	NamePos    scanner.Position
 	Value      Expression
 	OrigValue  Expression
-	Pos        scanner.Position
+	Equals     scanner.Position
```
danw: Maybe EqualsPos? Equals feels very non-specific to me. Colon below is a little better, but could still have a Pos suffix

colincross: Done

danw:
LGTM

PR #105, opened 2016-07-12
printer: support multiple skipped comments

colincross:
There may be multiple skipped comments in a row, convert
p.skippedComments to a list and add a test.

Change-Id: I30dcff269bee56fd51ef9513dab7c7885c44b7d7


danw:
LGTM

PR #107, opened 2016-07-25
Expose ModuleType to Singletons

danw:
In order to implement some build statistics in Soong, expose the module
type (the string used to define the module) to singletons.


colincross:
LGTM

PR #108, opened 2016-07-26
Prevent glob from accessing hidden files

danw:
Hidden files are often source control related (".git", ".repo", etc) or
editor related (vim's .*.swp), so are not guaranteed to exist, and may
be temporary. The build system shouldn't be using these files by
default.

If the glob pattern explicitly uses "." at the beginning of a path
component, allow returning hidden files for that component. Because of
this behavior, non-wildcard globs remain unchanged.

The one behavior that cannot be handled anymore is including hidden
files in recursive globs.

### DIFF
--- a/pathtools/glob.go
+++ b/pathtools/glob.go
@@ -131,6 +131,9 @@ func glob(pattern string, hasRecursive bool) (matches, dirs []string, err error)
 				if err != nil {
 					return nil, nil, err
 				}
+				if file[0] != '.' {
+					newMatches = filterDotFiles(newMatches)
+				}
 				matches = append(matches, newMatches...)
 			}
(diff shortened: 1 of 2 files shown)


colincross:
We already support GlobWithExcludes, why not have the primary builder provide the list of excludes it wants instead of enforcing .*?

danw:
For my usecase, this is required during the bootstrap phase with minibp, in order to use subdirs to read all the Android.bp files. So nothing from the primary builder is available.

In general, this behaves very similar to the normal shell commands (globbing, ls, etc) -- ignoring hidden files unless specifically asking for them.

colincross:
LGTM

PR #110, opened 2016-08-05
Optimize proptools

colincross:
Reduce primary builder times with two optimizations in proptools:
1.  Cache results of reflect.Type.Field() to avoid the garbage produced in reflect.StructField.Index
2.  Allow nil pointers to concrete struct types in properties to reduce allocating and copying empty structs

Also adds a -trace argument to the primary builder to allow tracing of concurrency.


danw:
LGTM

PR #111, opened 2016-08-09
Reimplement VisitDirectDeps[If] on Context.walkdeps

colincross:
VisitDirectDeps[If] need to know the parent in order to support
OtherModuleDependencyTag(), remove Context.visitDirectDeps[If] and
reimplement [Module]Context.VisitDirectDeps[If] using Context.walkdeps.

Also makes testing easier by mocking the filesystem and adds tests for VisitDirectDeps[If].


danw:
LGTM

PR #112, opened 2016-08-09
Only instantiate nil struct pointers that are used

colincross:
unpackProperties was instantiating all nil struct pointers in any struct
that had a property set.  Check if the property is set first, and only
instantiate the struct if it was set.

This has a slight behavioral change, as now structures that only exist
through nil pointers and are never set into will not be type-checked.
This should be fixed in a later patch set that moves the type checking
to be done across all property structs once per factory before blueprint
files are loaded.

Change-Id: I0dea34d7fff76bb4fc907516a2d996e4ea2408d6


danw:
LGTM

PR #113, opened 2016-08-10
Add concurrency to BottomUpMutators and cloneModules

colincross:
Profiling shows BottomUpMutators and cloneModules as the slowest parts of Soong, primarily due to lack of parallelism.  Allow primary builders to request BottomUpMutators to run in parallel, and always parallelize cloneModules.


danw:
LGTM

PR #115, opened 2016-08-11
Parallelize TopDownMutators

colincross:
Convert parallelVisitAllBottomUp into a generic parallelVisit that takes
a visitOrderer to select top down or bottom up.  Combine
runTopDownMutator and runBottomUpMutator into runMutators that takes a
mutatorDirection to select the visitOrderer for parallelVisit and which
function pointer in the mutatorInfo to run.  Optimize out the
updateDependencies to only run if dependencies have been modified to
avoid running it after TopDownMutators that cannot modify dependencies.

Change-Id: Ib00302db1108ebab2ce8e01b20aa026140d382a4

### DIFF
--- a/context.go
+++ b/context.go
@@ -74,6 +74,8 @@ type Context struct {
 	variantMutatorNames []string
 	moduleNinjaNames    map[string]*moduleGroup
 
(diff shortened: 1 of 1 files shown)


danw:
LGTM

PR #116, opened 2016-08-11
Fix error reporting in parallel mutators

colincross:
runMutators was sending the wrong value into errsCh, hiding all errors.
parallelVisit had a data race on error, send the cancel signal over a
channel instead of writing to a bool from a goroutine.

Change-Id: I18919d71464c094db5846badbfac80f99366d70c


danw:
LGTM

PR #117, opened 2016-08-12
Elide empty variations

colincross:
Don't add a "_" to the variant name if the variation name is empty.  For
example, when splitting a variant with name "foo" into variations "" and
"bar", the new variants would be named "foo" and "foo_bar" instead of
"foo_" and "foo_bar".

Change-Id: I82342d57e2a8e9f2d65a7d8d2872dcb7b3512899


danw:
I'm slightly worried that two mutators with the same variation names cause conflicts due to the elision. But hopefully the variations are named differently enough 😄 

LGTM

PR #119, opened 2016-08-22
Relax type requirements when extending properties

colincross:
Allow using ExtendMatchingProperties to extend pointer to a struct or an
interface containing a pointer to a struct using a struct, and
vice-versa.

Also fixes a pre-existing bug where extending a nested structure could
fail if there were multiple possible destnations and some of them did
not have a matching nested property.

Change-Id: I6e69d78eb6595ba7dd2603e3aa7dd8de3f292744


danw:
LGTM

PR #120, opened 2016-08-29
Add NewBuiltinRule and NewBuiltinPool

colincross:
Allow Blueprint build statements and rules to reference rules and pools
defined outside of Blueprint.

Change-Id: I48644497d16a4c2bab88db400dbe7f7fe169d933


danw:
LGTM

PR #128, opened 2016-10-12
Modify module name handling

colincross:
This patch set allows primary builders to have control over the name visible to Blueprint, to determine if a module with a name exists, to rename a module during a mutator, and to replace dependencies on a variant of module with another module.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -192,7 +192,9 @@ func isBootstrapBinaryModule(module blueprint.Module) bool {
 
 // A goPackage is a module for building Go packages.
 type goPackage struct {
+	blueprint.SimpleName
 	properties struct {
+		Deps      []string
 		PkgPath   string
 		Srcs      []string
 		TestSrcs  []string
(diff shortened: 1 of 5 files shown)


context.go
```diff
@@ -2191,6 +2221,18 @@ func (c *Context) handleRenames() []error {
 		group.name = name
 	}
 
+	for _, replace := range c.replacements {
+		// TODO(ccross): this probably has data races
```
danw: Why? this runs after the mutator pass, not during

context.go
```diff
@@ -2191,6 +2221,18 @@ func (c *Context) handleRenames() []error {
...
+		for _, m := range replace.from.reverseDeps {
+			for i, d := range m.directDeps {
+				if d.module == replace.from {
+					m.directDeps[i].module = replace.to
+				}
+			}
+		}
+
+		atomic.AddUint32(&c.depsModified, 1)
```
danw: This can be c.depsModified++, but this is also ignored, since we clear and check depsModified within runMutator, which is called before this function (this would work if we were called by runMutator)

colincross: Moved to runMutator

context.go
colincross: Oops, left over from when I was handling replacements immediately

context.go
```diff
@@ -2144,7 +2153,46 @@ func (c *Context) walkDeps(topModule *moduleInfo,
...
+		group, name := rename.group, rename.name
+		if name == group.name {
+			continue
+		}
+
+		existing := c.moduleNames[name]
+		if existing != nil {
+			errs = append(errs,
+				&BlueprintError{
+					Err: fmt.Errorf("renaming module %q to %q conflict with existing module",
```
colincross: Done

bootstrap/bootstrap.go
```diff
@@ -193,6 +193,8 @@ func isBootstrapBinaryModule(module blueprint.Module) bool {
 // A goPackage is a module for building Go packages.
 type goPackage struct {
 	properties struct {
+		Name      string
```
colincross: Done

PR #130, opened 2016-10-26
Support implicit outputs

danw:
Added in Ninja 1.7, for outputs that will not show up in $out.

(Also upgrades travis to ninja 1.7.1)


colincross:
LGTM

PR #133, opened 2016-11-05
Add better error message when stat returns nil

colincross:
Sometimes os.Stat on a path seems to return nil after filepath.Glob
returned the path as valid.  Return the error message to see why.

Bug: 32676828
Test: builds
Change-Id: Ieedddb673b4d641e5de08778febeb3d8ea025c0d

### DIFF
--- a/pathtools/glob.go
+++ b/pathtools/glob.go
@@ -16,6 +16,7 @@ package pathtools
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pathtools/glob.go
```diff
@@ -113,7 +114,11 @@ func glob(pattern string, hasRecursive bool) (matches, dirs []string, err error)
 	}
 
 	for _, m := range dirMatches {
-		if info, _ := os.Stat(m); info.IsDir() {
+		info, err := os.Stat(m)
+		if err != nil {
+			return nil, nil, fmt.Errorf("unexpected error after stat: %s", err)
```
colincross: I meant "unexpected error after glob"
os.Stat returns an os.PathError which will print:
stat: <path>: <error>

PR #141, opened 2017-02-03
Add globbing to filesystem mocking

colincross:
Add globbing to filesystem mocking so that more code can be tested
against the mock.  Also moves the filesystem mock to pathtools,
and renames pathtools.GlobWithExcludes to pathtools.Glob, replacing
the existing pathtools.Glob.

Test: blueprint tests
Change-Id: I722df8121bc870c4a861d7c249245c57dcb607be


pathtools/glob.go
```diff
@@ -52,11 +38,15 @@ func Glob(pattern string) (matches, dirs []string, err error) {
 // In general ModuleContext.GlobWithDeps or SingletonContext.GlobWithDeps
```
danw: Another copy of GlobWithExcludes two lines up (but github won't let me comment on those)

PR #143, opened 2017-02-08
Some progress on cleanup of the parser by removing Pos

mathjeff:
This is only a draft and doesn't quite work exactly right yet
It exists to facilitate getting some high-level comments

Change-Id: Id16b64670b012da8e045f443d846a9f9b6453675

### DIFF
--- a/bpfmt/bpfmt.go
+++ b/bpfmt/bpfmt.go
@@ -58,7 +58,7 @@ func processFile(filename string, in io.Reader, out io.Writer) error {
 
 	r := bytes.NewBuffer(src)
 
-	file, errs := parser.Parse(filename, r, parser.NewScope(nil))
+	file, errs := parser.Parse(filename, r, parser.NewScope(nil), false, false)
 	if len(errs) > 0 {
 		for _, err := range errs {
 			fmt.Fprintln(os.Stderr, err)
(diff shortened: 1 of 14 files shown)


mathjeff:
I think the summary of the state of this at the moment is:

This adds a builder class in the parser for programmatic construction of a tree (although at the momement it doesn't do much other than validate that the tree makes sense (which currently means making sure that comments aren't attached to nonexistent nodes))

This removes most of the instances of Pos, so we don't have to pass NoPos around in cases where it doesn't make sense. It should be easier to move pieces of the tree around now and have their comments automatically move with them. I still have to re-add it in a few places so we can still get line numbers inside of error messages. I'm thinking that'll probably entail putting a map on the SyntaxTree telling the original position for each ParseNode

I think that all of the parser and printer tests pass but I still have to make bpformat and android.mk compile and work again

This enables parser_test to explain where the differences are between the expected parse and the actual parse

It's still one large patch that might get broken up into smaller patches once the details are done

There's some other assorted cleanup that needs to be done involving like deleting some unused comments

If someone wants to take a short look at this, that'd be great

mathjeff:
Replacing this pull request with https://github.com/google/blueprint/pull/144

PR #145, opened 2017-03-09
Remove pos

mathjeff:


mathjeff:
androidmk patch is https://android-review.googlesource.com/c/333894/

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+		depth)
+}
+
+func comparePrimitives(actual namedItem, expected namedItem, depth int) (equal bool, diff string) {
+	// TODO make this less hacky
+	var a, b interface{}
+	if actual.item.Kind() != expected.item.Kind() {
+		return false, fmt.Sprintf("%s is of type %s whereas %s is of type %s", actual.name, actual.item.Kind(), expected.name, expected.item.Kind())
+	}
+	switch actual.item.Kind() {
```
colincross: This conversion is unnecessary, use
`reflect.DeepEqual(actual.item.Interface(), expected.item.Interface())`

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+
+}
+
+func deepCompareMaps(actual namedItem, expected namedItem, depth int) (equal bool, diff string) {
+	a := actual.item
+	b := expected.item
+	aKeys := a.MapKeys()
+	bKeys := b.MapKeys()
+	for _, aKey := range aKeys {
+		if !(isIn(aKey, bKeys)) {
```
colincross: This is unnecessary, use:
if !b.MapIndex(aKey).IsValid()

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
-					t.Errorf("test case: %s", testCase.input)
-					t.Errorf("incorrect comment %d:", i)
-					t.Errorf("  expected: %s", testCase.comments[i])
-					t.Errorf("       got: %s", file.Comments[i])
-				}
+	}
+	return item, false
+}
+
+func isIn(item reflect.Value, list []reflect.Value) (contained bool) {
```
colincross: I think this function (and findMatching and compareValues) are unnecessary due to comments below

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+		}
+	}
+	for _, bKey := range bKeys {
+		if !(isIn(bKey, aKeys)) {
+			bValue := b.MapIndex(bKey)
+			return false, fmt.Sprintf("%v contains key %s (with value %#v) but %v does not", expected.name, bKey, bValue, actual.name)
+		}
+	}
+	for _, aKey := range aKeys {
+		bKey, _ := findMatching(aKey, bKeys)
```
colincross: This seems unnecessary, findMatching is going to return a key that is deeply equal to aKey, so why not just use aKey?

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+	}
+	return true, ""
+}
+
+func deepCompareArrays(actual namedItem, expected namedItem, depth int) (equal bool, diff string) {
+	a := actual.item
+	b := expected.item
+	aLen := a.Len()
+	bLen := b.Len()
+	sharedLen := int(math.Min(float64(aLen), float64(bLen)))
```
colincross: Don't cast to float, either implement a trivial min function or open code it

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+	for i := 0; i < sharedLen; i++ {
```
colincross: Or: 
`for i := 0; i < a.Len() && i < b.Len(); i++ {`

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+	}
+	return true, ""
+}
+
+func deepCompareStructs(actual namedItem, expected namedItem, depth int) (equal bool, diff string) {
+	a := actual.item
+	b := expected.item
+	aCount := a.NumField()
+	bCount := b.NumField()
+	if aCount != bCount {
```
colincross: It would be more reliable to compare the types here, or delete the check entirely

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+			namedItem{expected.name + "." + fieldName, b.Field(i)},
+			depth)
+		if !equal {
+			return equal, diff
 		}
 	}
+	return true, ""
+}
+
+// deepCompare is much like reflect.DeepEqual, but it does a deep comparison of map keys too (which is slower than a map lookup)
```
colincross: Deep comparing map keys seems unnecessary to me.  If they hash to the same value they must be == equal, and if they are == equal then they must be DeepEqual.  If that is not true, can you construct a test that returns different values for reflect.DeepEqual and deepCompare?

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+// also, deepCompare gives an explanation of the first difference that it finds
+func deepCompare(actual interface{}, expected interface{}) (equal bool, diff string) {
+	return deepCompareObjects(namedItem{"actual", reflect.ValueOf(actual)}, namedItem{"expected", reflect.ValueOf(expected)}, 0)
+}
+
+// deprecated
```
colincross: Why check in a deprecated function?

parser/parser.go
```diff
@@ -37,77 +37,193 @@ func (e *ParseError) Error() string {
...
-	errs = p.errors
-	comments := p.comments
-
-	return &File{
-		Name:     p.scanner.Filename,
-		Defs:     defs,
-		Comments: comments,
-	}, errs
-
+type parser struct {
```
colincross: reordering these makes this really hard to review

parser/parser.go
```diff
@@ -37,77 +37,193 @@ func (e *ParseError) Error() string {
 	return fmt.Sprintf("%s: %s", e.Pos, e.Err)
 }
 
-type File struct {
-	Name     string
-	Defs     []Definition
-	Comments []*CommentGroup
+type ParseTree struct {
```
colincross: The distinction between ParseTree and SyntaxTree is not clear to me.  Leaving this as File and File.Name would be much easier to review, and then you can use a separate patch to do renames.  The gorename command line tool is useful for doing programmatic renames, and you can include the gorename command in the commit message.

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
 			t.Errorf("test case: %s", testCase.input)
 			t.Errorf("length mismatch, expected %d definitions, got %d",
 				len(testCase.defs), len(file.Defs))
+
```
mathjeff: good idea

parser/parser_test.go
mathjeff: I'll remove it

parser/parser_test.go
mathjeff: ok

parser/parser_test.go
mathjeff: I'll re-implement the 'min' function as you mentioned above.

I extracted the 'min' operation to slightly decrease the number of values being recomputed during each loop iteration (I don't see anything in https://github.com/golang/go/wiki/CompilerOptimizations specifying that the Go compiler can optimize 'i < a.Len() && i < b.Len()').

parser/parser_test.go
mathjeff: ok

parser/parser_test.go
mathjeff: Yeah, I was doing a direct map lookup initially. Since the keys are themselves pointers pointing to different but recursively equal objects, the lookup wasn't finding the match.

Technically there's a detail here that doesn't get tested directly by the test now, which is that the key in the map not only should be recursively equal to another map key in the original object, but also that it should be the same reference as another object in the parse tree somewhere. For the moment that check is in the builder class itself though.

The more that I work on this parser test, the more I wonder if it would be better to have the parser test use the printer and do a text diff instead, and then to reimplement the printer test to directly pass parseTrees into the printer and to not have it use the parser at all. I guess my default is to not bother to rearrange the parser and printer tests, but I could do that too if you think it would be better. Technically that approach wouldn't be able to rule out the possibility that the parser was outputting some unexpected data structure that the printer just so happened to accept and just so happened to print in a way that matched the desired output - I hypothesize we wouldn't be very concerned about that possibility either.

parser/parser_test.go
mathjeff: See below about keys being pointers

parser/parser_test.go
```diff
@@ -639,23 +639,276 @@ func TestParseValidInput(t *testing.T) {
...
+	}
+	return item, false
+}
+
+func isIn(item reflect.Value, list []reflect.Value) (contained bool) {
+	_, found := findMatching(item, list)
+	return found
+}
+
+// checks if two objects are the same class, returns True if they match
```
mathjeff: ok

parser/parser_test.go
mathjeff: cool

parser/parser_test.go
mathjeff: See below about keys being pointers

parser/parser_test.go
mathjeff: Yeah, I was surprised too. I'll add a test involving a map having keys of pointer type.

parser/parser_test.go
colincross: What happens if two deeply equal pointers are keys in the same map?  Only the pointer value will be hashed, so they will still be two different keys.

mathjeff: Yeah exactly; the test doesn't confirm that the pointers themselves are correct, only that they have a corresponding value in the other map.

Do you mean that we want to fix deepCompare for the following case?

////
var a and b, deeply equal to each other
var c and d, deeply equal to each other

map1{a:true, b:true, c:true}
map2{a:true, c:true, d:true}

?

for which deepCompare would return true when really we want false

colincross: We would need to define what the semantics are for comparing two maps whose keys are deepequal but not equal.  Even for two maps that were deep-cloned off of each other, there is no guarantee that we can determine a 1-1 mapping between keys in the two maps.

The alternative is to sidestep this problem and refuse to deepcompare pointers or anything that might contain pointers (structs, arrays) as the keys in maps.  This should be fine for the parser/printer, since they only support strings as the keys to maps.

mathjeff: Unfortunately the test needs to be able to compare SyntaxTree.comments, whose type is map[ParseNode](*CommentPair) , which uses an interface type as its key

I think that my inclination, if we want to spend the effort, would be two changes. #1 we fix the case in deepCompare involving maps having recursively equal keys such that deepCompare deems them equal, and #2 we require the individual tests to explicitly invoke deepCompare, so if the tests require certain relationships between the pointers themselves (currently the checks specific to the pointers are in the Builder class itself), then the tests can validate those requirements separately, and can still use deepCompare while doing it.

The changes to deepCompare I imagine would be to have it assume that if there exists a 1-1 mapping for which we cannot find a relevant difference, then we deem the data structures equivalent. In other words: 1. For each combination of entries in map1 and map2, check whether those two entries are equal. 2. For each entry A in map1 that has K matches in map2, confirm that there are exactly K such entries in map1 that have the same set of matches in map2 3. If every condition in #2 yielded true, then return true.


As for sidestepping the problem, my favorite option is to invoke the printer (or perhaps VerbosePrinter) in printer_test.go and check the text output.


Alternatively we could adjust the code being tested attach an integer id to each ParseNode and to use that as the map key rather than the ParseNode reference itself.

parser/parser_test.go
mathjeff: Well I investigated a little further but it seems that unfortunately, x.Interface() fails if there were any private fields traversed while accessing x. My thought now is that it's preferable to have support for diffs of private fields if it requires reimplementing this method like this.

parser/parser.go
mathjeff: done

parser/parser.go
mathjeff: done

parser/printer.go
```diff
@@ -330,36 +330,60 @@ func (p *printer) flush() {
...
-func (p *printer) indent(i int) {
-	p.indentList = append(p.indentList, i)
+func NewVerbosePrinter(tree *SyntaxTree) (p *VerbosePrinter) {
+	return &VerbosePrinter{tree, bytes.Buffer{}, 0}
 }
 
-func (p *printer) unindent(pos scanner.Position) {
-	p.printEndOfLineCommentsBefore(pos)
-	p.indentList = p.indentList[0 : len(p.indentList)-1]
+func (p *VerbosePrinter) Print() string {
```
colincross: This is the only public method, and always used as NewVerbosePrinter(...).Print(), might as well make a VerbosePrint(...) method that delegates to an unexported verbosePrinter type.

parser/printer_test.go
```diff
@@ -315,34 +315,67 @@ test { // test
...
-		expected := testCase.output[1:]
+	for i, testCase := range validPrinterTestCases {
+
+		description := testCase.description
+
+		in := testCase.input[:]
+		if in[0] == '\n' {
+			in = in[1:]
+		} else {
+			panic(fmt.Sprintf("A newline at the beginning of testcase input is required, to make the test code easily readable. Offending test case '%s' starts with character: '%s'", description, string(in[0])))
```
colincross: This doesn't seem worth enforcing, just strip it if it is there.

proptools/compare.go
```diff
@@ -0,0 +1,248 @@
...
+func DeepCompare(a NamedItem, b NamedItem) (comparison Comparison) {
+	return deepCompareGeneric(toReflection(a), toReflection(b), 0)
+}
+
+type NamedItem struct {
+	Name string
+	Item interface{}
+}
+
+type Comparison struct {
```
colincross: a two-value return is more idiomatic than returning a struct

proptools/compare.go
```diff
@@ -0,0 +1,248 @@
...
+		panic(fmt.Sprintf("unrecognized types, %s (%#v) and %s (%#v)", actual.name, actual.item, expected.name, expected.item))
+	}
+	if reflect.DeepEqual(a, b) {
+		return equal
+	} else {
+		return different(fmt.Sprintf("%s = %#v whereas %s = %#v", actual.name, actual.item, expected.name, expected.item))
+	}
+}
+
+// compares two maps
```
colincross: document the semantics of this comparison, especially as compared to reflect.DeepEqual

proptools/compare.go
```diff
@@ -0,0 +1,248 @@
...
+		)
+		if !comparison.Equal {
+			return comparison
+		}
+
+	}
+	return equal
+}
+
+// compares two arrays
```
colincross: compares two arrays or slices

document that this compares [0,len()), and not [0,cap())

proptools/compare.go
```diff
@@ -0,0 +1,248 @@
...
+		return deepComparePointers(a, b, depth)
+	case reflect.Int, reflect.String, reflect.Bool:
+		return comparePrimitives(a, b, depth)
+	default:
+		panic(fmt.Sprintf("unrecognized type %v of value %#v\n", a.item.Kind(), a.item))
+
+	}
+}
+
+// compares two pointers in the same manner as reflect.deepEqual
```
colincross: compares two pointers or interfaces

colincross (commented):
Partial review of the first 2 commits

parser/parser_test.go
```diff
@@ -16,7 +16,7 @@ package parser
 
 import (
 	"bytes"
-	"reflect"
+	"github.com/google/blueprint/proptools"
```
mathjeff: Have you made any customizations to Gogland that automate things like line lengths and import ordering? If so, I wonder if we can check them in somewhere.

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
-			t.Errorf("length mismatch, expected %d comments, got %d",
-				len(testCase.comments), len(file.Comments))
+// Assuming that the two trees have equivalent structure, this function returns a map that translates nodes in <actual> to <expected>
+// The behavior is undefined if the two trees don't have equal structure between their non-comment nodes
+func getNodeMapping(expected *SyntaxTree, actual *SyntaxTree) (remappings map[ParseNode]ParseNode) {
+	actualNodes := actual.ListOfAllNodes()
+	expectedNodes := expected.ListOfAllNodes()
+	remappings = make(map[ParseNode]ParseNode, 0)
+	for i, actual := range actualNodes {
+		remappings[actual] = expectedNodes[i]
```
mathjeff: ok

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+
+	analogousActualComments := make(map[ParseNode](*CommentPair), 0)
+	for actualKey, actualValue := range actual.SyntaxTree.comments {
+		analogousKey := actualNode_remappings[actualKey]
+		analogousActualComments[analogousKey] = actualValue
+	}
+
+	// now compare the comments
+	return proptools.DeepCompare(
+		proptools.NamedItem{"expected.SyntaxTree.Comments", expected.SyntaxTree.comments},
```
mathjeff: ok

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+		proptools.NamedItem{"actual.SyntaxTree.Comments", analogousActualComments})
+}
+
+func compareSourcePositions(expected *ParseTree, actual *ParseTree, actualNode_remappings map[ParseNode]ParseNode) proptools.Comparison {
```
mathjeff: Yeah I don't like the underscore either; I just use it to clarify that this is the remappings of the actualNodes as opposed to being the actual nodeRemappings. What name would you recommend that conveys as much information?

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+	// Construct a new map of source positions, whose values are the same as <actual>, but whose keys are the corresponding keys in <expected>
+	// The reason we replace the keys is so proptools.DeepCompare can know which keys we want to compare against which others
+
+	analogousSourcePositions := make(map[ParseNode]scanner.Position)
+	for actualKey, actualValue := range expected.SourcePositions {
```
mathjeff: true

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+	if !nodesComparison.Equal {
+		return nodesComparison
+	}
+
+	// now that we've confirmed that the non-comment nodes have equivalent structure, we can compute a node mapping
+	remappings := getNodeMapping(expected.SyntaxTree, actual.SyntaxTree)
+
+	// use the node mapping to compare the comments
+	commentsComparison := compareComments(expected, actual, remappings)
+	if !commentsComparison.Equal {
```
mathjeff: ok

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+		return commentsComparison
+	}
+	return commentsComparison
+
+	// TODO update all the test cases to specify SourcePositions, and then compare the source positions too
+	// use the node mapping to compare the source positions
+	//return compareSourcePositions(expected, actual, remappings)
+}
+
+func runValidIndex(t *testing.T, i int) {
```
mathjeff: ok

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+	// TODO update all the test cases to specify SourcePositions, and then just use that
+	correctParse := &ParseTree{"correct output", correctTree, nil}
+
+	// confirm that the actual and expected trees are equivalent
+	comparison := compareParseTrees(actualParse, correctParse)
+
+	expectedRepresentation := NewVerbosePrinter(correctParse.SyntaxTree).Print()
+	actualRepresentation := NewVerbosePrinter(actualParse.SyntaxTree).Print()
+	if !comparison.Equal {
+		t.Errorf(`
```
mathjeff: ok

parser/parser_test.go
```diff
@@ -639,22 +639,123 @@ func TestParseValidInput(t *testing.T) {
...
+                %s
+expected:
+%s
+got     :
+%s
+1st diff: %v
+
+`, i, testCase.input, expectedRepresentation, actualRepresentation, comparison.Difference)
+	}
+	succeeded = true
```
mathjeff: thanks

parser/printer.go
```diff
@@ -330,36 +330,60 @@ func (p *printer) flush() {
 		p.curComment++
 	}
 	p.output = append(p.output, '\n')
+// a VerbosePrinter prints a syntax tree and is intended for debugging purposes
```
mathjeff: ok

parser/printer.go
mathjeff: ok

parser/printer_test.go
mathjeff: ok

proptools/compare.go
mathjeff: When is that (two values preferable to a struct) the case? Is it when the two values have fundamentally different meanings? It seems to me that a Comparison is itself a concept. If we wanted to instead return an entire list of differences, it'd be convenient to not have to update the signatures of each method based on the implementation of a Comparison.

proptools/compare.go
mathjeff: done

proptools/compare.go
mathjeff: done

proptools/compare.go
mathjeff: ok

proptools/compare.go
colincross: I'd say that a struct should be used if you are going to pass that encapsulated value around (more than just returning it).  If you are going to immediately unpack and operate on it, a multiple-value return is better.

parser/parser_test.go
colincross: Sometimes its better to use a short, readable name and put the extra information in a comment, rather than carrying all the information everywhere you use the name.  See https://github.com/golang/go/wiki/CodeReviewComments#variable-names

mathjeff: Hmm it seems to me that the trouble in this case is more that I forgot to add a comment than that I used a long variable name.

In any case, I've now redone it in the next version of this commit for you to take a look at.

proptools/compare.go
mathjeff: That logic sounds like a decent rule of thumb to me: to use a multi-valued return if the caller is likely to immediately read all of the contained fields. Does it apply here? None of the functions in compare.go read the field Comparison.Difference .

If for example we wanted to add another field to Comparison that perhaps specified the property path of the difference, then those fields probably wouldn't be read by anything in compare.go either.

mathjeff: I talked to Dan and he thinks that if the number of fields is small then the thing to do is to make the fields be separate return values, and if the number of fields grows then to consider making the return be a struct at that time. I suppose that that's enough of a rule of thumb for me to be able to follow anyway, so I'll do that.

parser/parser_test.go
colincross: No, and it keeps screwing them up for me too.

proptools/compare.go
```diff
@@ -0,0 +1,266 @@
...
+
+func same() (equal bool, difference string) {
+	return true, ""
+}
+
+func different(explanation string) (equal bool, difference string) {
+	return false, explanation
+}
+
+// end of public members
```
colincross: This seems both unnecessary and misleading, since same() and different() are not public

proptools/compare.go
```diff
@@ -0,0 +1,266 @@
...
+		return firstDifference(differences)
+	}
+	for _, bKey := range bKeys {
+		if !aItem.MapIndex(bKey).IsValid() {
+			bValue := bItem.MapIndex(bKey)
+			differences = append(differences, fmt.Sprintf("%v contains key %#v (with value %#v) but %v does not", b.name, bKey, bValue, a.name))
+		}
+	}
+	if len(differences) > 0 {
+		return firstDifference(differences)
```
colincross: return false, sort.Strings(differences)[0]

proptools/compare.go
```diff
@@ -0,0 +1,266 @@
...
+			differences = append(differences, difference)
+		}
+	}
+	if len(differences) > 0 {
+		return firstDifference(differences)
+	}
+	return same()
+}
+
+// compares two arrays (this ignores cap(), and only compares elements from indices 0 to len() )
```
colincross: arrays or slices

proptools/compare_test.go
```diff
@@ -0,0 +1,229 @@
...
+	{
+		map[int]string{0: "zero", 1: "one"},
+		map[int]string{0: "zero", 1: "not one"},
+		`a[1(%!t(int=1))] = "one" whereas b[1(%!t(int=1))] = "not one"`,
+	},
+	{
+		// a map with several differences, for confirming determinism
+		map[int]int{0: 0, 1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6, 7: 7, 8: 8, 9: 9},
+		map[int]int{0: 1, 1: 2, 2: 3, 3: 4, 4: 5, 5: 6, 6: 7, 7: 8, 8: 9, 9: 0},
+		`a[0(%!t(int=0))] = 0 whereas b[0(%!t(int=0))] = 1`,
```
colincross: This seems like a bug?

mathjeff: What would you prefer instead?

proptools/compare.go
mathjeff: oops

proptools/compare.go
mathjeff: thanks

parser/printer.go
```diff
@@ -15,351 +15,472 @@
...
-	if pos == noPos {
-		pos = p.pos
-	}
-
-	if newline {
-		p.printEndOfLineCommentsBefore(pos)
-		p.requestNewlinesForPos(pos)
+func (p *printer) beforePrint(node ParseNode) {
+	var comments = p.syntaxTree.GetCommentsIfPresent(node)
+	if comments != nil {
```
mathjeff: Hmm, when I remove the check, I get a panic trying to access comments.preComments

PR #149, opened 2017-03-20
Improve error messages when dependency variants are missing

danw:
Instead of just saying:

```
error: .../Android.bp:48:1: dependency "libc++" of "libtest" missing variant "arch:android_arm_armv7-a, link:shared, vndk:"
```

Include a list of currently existing variants:

```
error: .../Android.bp:48:1: dependency "libc++" of "libtest" missing variant:
  arch:android_arm_armv7-a, link:shared, vndk:
available variants:
  arch:android_arm_armv7-a, link:shared
  arch:linux_x86, link:shared
  arch:linux_x86_64, link:shared
  arch:windows_x86, link:shared
  arch:windows_x86_64, link:shared
```

This still isn't the best experience for users, but it at least provides enough information for someone more familiar with the build to understand the problem.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1227,10 +1227,17 @@ func (c *Context) addDependency(module *moduleInfo, tag DependencyTag, depName s
 		return nil
 	}
 
+	variants := make([]string, len(possibleDeps))
+	for i, mod := range possibleDeps {
+		variants[i] = c.prettyPrintVariant(mod.variant)
+	}
+	sort.Strings(variants)
+
 	return []error{&BlueprintError{
-		Err: fmt.Errorf("dependency %q of %q missing variant %q",
+		Err: fmt.Errorf("dependency %q of %q missing variant:\n  %s\navailable variants:\n  %s",
 			depName, module.Name(),
-			c.prettyPrintVariant(module.dependencyVariant)),
+			c.prettyPrintVariant(module.dependencyVariant),
+			strings.Join(variants, "\n  ")),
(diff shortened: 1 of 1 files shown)


danw:
Colin, does this look interesting for normal use, or should I just keep it around for debugging?

context.go
```diff
@@ -1256,10 +1263,17 @@ func (c *Context) findReverseDependency(module *moduleInfo, destName string) (*m
...
 
+	variants := make([]string, len(possibleDeps))
+	for i, mod := range possibleDeps {
+		variants[i] = c.prettyPrintVariant(mod.variant)
+	}
+	sort.Strings(variants)
+
 	return nil, []error{&BlueprintError{
-		Err: fmt.Errorf("reverse dependency %q of %q missing variant %q",
+		Err: fmt.Errorf("reverse dependency %q of %q missing variant:\n  %s\navailable variants:\n  %s",
```
Nan-Zhang: is there specific reasons that we change variant format from %q to %s?

danw: When they're on their own line, it doesn't feel like they need to be quoted

mathjeff: How does a reverse dependency differ from a non-reverse dependency? Does this message appear to users?

colincross: reverse dependency is rarely used, when a module wants to add itself as a dependency of another module instead of vice-versa.  It is used by prebuilts, so yes, it can appear to users.

PR #150, opened 2017-03-21
Clarify bootstrap documentation/usage

mathjeff:
Change-Id: Idbf80b9f5d1a3c7b47db8dad83df550612916958


bootstrap.bash
```diff
@@ -1,13 +1,15 @@
...
+# run this script with no arguments. The generated files will be placed in the
+# working directory (unless BUILDDIR is set to something else).
 #
-# It can also be invoked from another script to bootstrap a custom Blueprint-
-# based build system.  To do this, the invoking script must first set some or
-# all of the following environment variables, which are documented below where
-# their default values are set:
+# This script can also generate a custom bootstrap manifest (most commonly
+# called "build.ninja.in") for use in another Blueprint-based build system.
+# This is activated by the '-r' command-line option.
```
danw: The -r option would actually be the third purpose (and we'd use it to generate the build.ninja.in in this directory too, not just for other Blueprint-based systems.

bootstrap.bash
```diff
@@ -115,17 +114,19 @@ while getopts ":b:hi:rt" opt; do
...
-        echo "Regenerating $BOOTSTRAP_MANIFEST"
         $BUILDDIR/.bootstrap/bin/minibp $EXTRA_ARGS -o $BOOTSTRAP_MANIFEST $SRCDIR/$TOPNAME
     else
-        echo "Executable minibp not found at $BUILDDIR/.bootstrap/bin/minibp" >&2
+        echo "Executable minibp not found at $BUILDDIR/.bootstrap/bin/minibp. Ensure that the environment variable BUILDDIR refers to the output directory of a previous successful build of Blueprint." >&2
         exit 1
     fi
 fi
 
+echo "Generating bootstrap files at $BUILDDIR"
```
danw: We wouldn't want this line at the beginning of every Android build.

bootstrap.bash
```diff
@@ -1,13 +1,15 @@
 #!/bin/bash
 
-# This script serves two purposes.  First, it can bootstrap the standalone
-# Blueprint to generate the minibp binary.  To do this simply run the script
-# with no arguments from the desired build directory.
+# This script serves two purposes.  First, it can generate bootstrapping files
```
Nan-Zhang: What the specific bootstrapping files are generated? is it the ninja file?

mathjeff: As far as I can tell that's essentially right. I've updated the comment accordingly.

Although part of my reason for making this pull request was because I wasn't fully sure of how things worked, and this way people would correct me if my clarifications were incorrect

PR #151, opened 2017-03-30
Added util functionalities to return build actions to Soong layer.

Nan-Zhang:
Added a public build-actions struct which can be used to compared with expected build actions data in Soong layer.

Added a public util function so that we can choose specific module build actions to be compared in Soong testing/debugging mode.

### DIFF
--- a/context.go
+++ b/context.go
@@ -204,6 +204,70 @@ func (module *moduleInfo) String() string {
 	return s
 }
 
+func (module *moduleInfo) GetBuildActions(c *Context) NinjaBuildActions {
+	ret := NinjaBuildActions{}
+
+	for _, v := range module.actionDefs.variables {
+		name := v.fullName(nil)
+		value, err := v.value(nil)
+		if err != nil {
+			panic(err)
+		}
+		ret.Variables = append(ret.Variables,
+			NinjaVariable{Name: name, Value: value.Value(c.pkgNames)})
+	}
+
+	for _, r := range module.actionDefs.rules {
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -2535,13 +2577,32 @@ func (c *Context) ModuleErrorf(logicModule Module, format string,
...
+		}
+	})
+}
+
+func (c *Context) GetModuleBuildActionsIf(pred func(Module) bool) NinjaBuildActions {
+	var ret NinjaBuildActions
+
+	c.visitAllModules(func(module *moduleInfo) {
+		if pred(module.logicModule) {
+			ret = module.GetBuildActions(c)
```
mathjeff: Do we want to have some sort of validation that <ret> only ever gets set once?

context.go
```diff
@@ -204,6 +204,70 @@ func (module *moduleInfo) String() string {
 	return s
 }
 
+func (module *moduleInfo) GetBuildActions(c *Context) NinjaBuildActions {
```
mathjeff: Will anything in Blueprint call this, or is it intended for the Soong unit tests?

Basically I wonder if we can refactor something Blueprint to use this to enforce that it remains up-to-date even if things change. If we have time, my inclination would be for ninja_writer.go#Build to take a NinjaBuildActions as an argument. It looks like ninja_defs.go might be related too.

Nan-Zhang:
Since Colin has provided a patch for adding integration test, abandoned this patch so.

PR #152, opened 2017-04-10
Added OtherModuleType() util function.

Nan-Zhang:
Use case:

A Python binary should not be depended by other modules.

So during the process that a specific module is visiting all dependencies:

ctx.VisitDepsDepthFirst(func(module blueprint.Module) {
     if ctx.OtherModuleType(module) == "python_binary_host" {
			panic(fmt.Errorf(
				"module %q directly or indirectly depends on Python binary: %q.",
				ctx.ModuleName(), ctx.OtherModuleName(module)))
		}
                ......
}

### DIFF
--- a/module_ctx.go
+++ b/module_ctx.go
@@ -150,6 +150,7 @@ type ModuleContext interface {
 	OtherModuleName(m Module) string
 	OtherModuleErrorf(m Module, fmt string, args ...interface{})
 	OtherModuleDependencyTag(m Module) DependencyTag
+	OtherModuleType(m Module) string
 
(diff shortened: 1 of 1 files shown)


colincross:
You can do this with something like:
```go
ctx.VisitDepsDepthFirst(func(module blueprint.Module) {
    if _, ok := module.(*pyBinary); ok {
        panic(fmt.Errorf("module %q directly or indirectly depends on Python binary: %q.",
            ctx.ModuleName(), ctx.OtherModuleName(module)))
    }
}
```

Nan-Zhang:
You are right. Thanks Colin.

PR #154, opened 2017-05-16
Cap concurrency when parsing blueprint files

colincross:
Darwin has a default limit of 256 open files per process.  Parsing
too many blueprint files in parallel can hit the limit.  Cap the
concurrency at 200.

Test: manual testing with limit set to 32
Change-Id: Ic64d21d2c0ffd7c86bf3f02fb51216ee5684a80c

### DIFF
--- a/context.go
+++ b/context.go
@@ -719,10 +719,14 @@ func (c *Context) WalkBlueprintsFiles(rootFile string, handler FileHandler) (dep
 	// Number of outstanding goroutines to wait for
 	count := 0
 
-	startParseBlueprintsFile := func(filename string, scope *parser.Scope) {
+	startParseBlueprintsFile := func(blueprint stringAndScope) {
+		if blueprintsSet[blueprint.string] {
+			return
+		}
+		blueprintsSet[blueprint.string] = true
 		count++
 		go func() {
-			c.parseBlueprintsFile(filename, scope, rootDir,
+			c.parseBlueprintsFile(blueprint.string, blueprint.Scope, rootDir,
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -749,14 +755,19 @@ loop:
...
 			}
-
-			blueprintsSet[blueprint.string] = true
-			startParseBlueprintsFile(blueprint.string, blueprint.Scope)
+			startParseBlueprintsFile(blueprint)
 		case <-doneCh:
 			count--
+			if len(pending) > 0 {
+				startParseBlueprintsFile(pending[len(pending)-1])
+				pending = pending[0 : len(pending)-1]
```
Nan-Zhang: pending = pending[:len(pending) - 1]

context.go
```diff
@@ -749,14 +755,19 @@ loop:
 			if tooManyErrors {
 				continue
 			}
-			if blueprintsSet[blueprint.string] {
+			// Limit concurrent calls to parseBlueprintFiles to 200
+			// Darwin has a default limit of 256 open files
```
danw: Any idea if this limits our performance? should we either auto-detect this, or have separate mac vs linux limits?

context.go
colincross: Done

context.go
colincross: I wouldn't think so - we want at least a few more goroutines than the number of cpus to allow for blocking on IO, but after that I'd expect adding more goroutines to reduce performance, not increase it.  200 is still probably more than we need.

PR #156, opened 2017-06-23
Fix glob deps for non-wild patterns

danw:
Patterns that were not wild would return an empty "dirs" list if the
file was found. But then if they were removed, we wouldn't know to
update the glob file and re-run the primary builder.

In this case, instead of adding the final directory into the dirs list,
add the matched files themselves. Due to editors performing atomic
writes (the directory timestamp often gets updated at the same time as
file timestamp) this is probably more efficient. In either case, we're
only re-running the individual glob, which is rather cheap.

Rename startGlob/Glob return name from "dirs" to "deps" since it may
contain files now too.

### DIFF
--- a/pathtools/glob.go
+++ b/pathtools/glob.go
@@ -28,25 +28,25 @@ import (
 var GlobMultipleRecursiveErr = errors.New("pattern contains multiple **")
 var GlobLastRecursiveErr = errors.New("pattern ** as last path element")
 
-// Glob returns the list of files that match the given pattern but
-// do not match the given exclude patterns, along with the list of directories
-// that were searched to construct the file list.  The supported glob and
-// exclude patterns are equivalent to filepath.Glob, with an extension that
-// recursive glob (** matching zero or more complete path entries) is supported.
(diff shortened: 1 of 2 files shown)


pathtools/glob.go
```diff
@@ -58,7 +58,11 @@ func startGlob(fs FileSystem, pattern string, excludes []string) (matches, dirs
 		return nil, nil, err
 	}
 
-	return matches, dirs, nil
+	if !isWild(pattern) {
```
colincross: A comment here would help.

pathtools/glob_test.go
```diff
@@ -110,12 +110,12 @@ var globTestCases = []struct {
 	{
 		pattern: "a",
 		matches: []string{"a"},
-		dirs:    nil,
+		dirs:    []string{"a"},
```
colincross: Rename this to deps too?

danw: Done

pathtools/glob.go
danw: Done

PR #162, opened 2017-07-17
Unify build directories

danw:
Having separate .ninja_log files doesn't work very well after
https://github.com/ninja-build/ninja/commit/04d886b11041bb59d01df794cce7a1e8cad2250d
since the last successful timestamp is now recorded in the ninja log.
This was triggering the primary builder to be run twice on every build
since it was always considered out of date in the primary, then main
stages.

So always use the same build directory so that we share a .ninja_log.
This means that we can turn off the "generator" option to the primary
builder in the main stage, since it'll have the correct command line
that we ran in the primary stage. We also save some time from not
running globs shared between minibp and the primary builder twice.

The cleanup code is removed, since we no longer know what was built in
the different stages from just reading the ninja log, and we can't turn
it on for the main stage.

Change-Id: Ib036123139621f88ee0cf546ccded27006ff1410


danw:
Yeah, I'm a bit worried about parsing the much larger .ninja_log files in those cases though.

danw:
And this doesn't actually fix the problem with Soong, since Soong is resetting the ninja build directory (to out from out/soong)

PR #163, opened 2017-07-21
Fuller MockFs to facilitate tests of the cacheable finder

mathjeff:
Bug: 62455338
Test: m -j blueprint_tools # which runs the unit tests

Change-Id: I5ab2650459a1dae0d5d076faf411ec2d053c743d

### DIFF
--- a/context.go
+++ b/context.go
@@ -780,7 +780,7 @@ loop:
 // MockFileSystem causes the Context to replace all reads with accesses to the provided map of
 // filenames to contents stored as a byte slice.
 func (c *Context) MockFileSystem(files map[string][]byte) {
-	c.fs = pathtools.MockFs(files)
+	c.fs = pathtools.NewMockFs(files)
 }
 
 // parseBlueprintFile parses a single Blueprints file, returning any errors through
--- a/pathtools/fs.go
+++ b/pathtools/fs.go
@@ -20,47 +20,53 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/user"
 	"path/filepath"
+	"sync/atomic"
+	"syscall"
+	"time"
 )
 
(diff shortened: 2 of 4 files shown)


colincross:
The MockFs that finder needs is completely disjoint from what blueprint needs.  I think you should create your own MockFs alongside finder that serves your needs without complicating the one used by blueprint.

mathjeff:
So we're not concerned that splitting the mock file systems means that if something else (most likely another test) needs a mock filesystem, then there would be two mock filesystems to choose from, neither of which having the full feature set?

colincross:
We can reconcile them later if it turns out both are needed, but I don't expect many things in the system to care about timestamps, since that is normally ninja's job.

mathjeff:
What's simpler about splitting MockFs into two? It seems to me that the primary advantage of splitting it is instead a decrease in risk - it prevents a MockFs change relating to Finder from breaking Glob. However, that seems to me to instead be a reason to add dedicated tests of MockFs.

This patch also adds the ability to modify the filesystem (this version adds the ability to rename a file; I also implemented deletion and recursive deletion but haven't uploaded it yet). I also noticed recently that I'm going to want to add support for symlinks and for file/directory permissions too.

colincross:
Mostly keeping the complexity out of blueprint if it is unnecessary.  The MockFs in blueprint is less than 100 lines, and mostly trivial, and is used to test some fairly complex code in Glob.  You are adding 400 more lines that blueprint doesn't need, and your test doesn't need much of the existing code.

mathjeff:
Oh. Is the reason to want to keep complexity out of the MockFs from blueprint/pathtools/fs.go because it's used in several places via `ctx.MockFileSystem`?

mathjeff:
(that seems like a good reason to me; just clarifying what it was that I didn't know previously)

PR #164, opened 2017-07-24
Simplify bootstrapping with microfactory

danw:
This simplifies our bootstrapping procedure by removing build.ninja.in and all the logic required to create it. Instead, we build minibp using microfactory, which requires almost no configuration to build Go tools. Then we use a 19 line ninja file to run minibp and create .bootstrap/build.ninja like before.

We've been using microfactory for a while in Soong to build soong_ui, which is the Go-based build wrapper. It's slightly slower than the equivalent ninja file, but still only takes ~20ms if there is no work to be done.

Besides the simplification, the other goal of this pull request is to fix the issue caused by https://github.com/ninja-build/ninja/commit/04d886b11041bb59d01df794cce7a1e8cad2250d.  We can no longer build the same file in two different ninja runs that use different .ninja_log files, otherwise we'll build that file in both stages on every build, even when it's up to date.


danw:
The Soong side of this change: https://android-review.googlesource.com/c/442531/

PR #165, opened 2017-07-25
Go 1.9 changes

danw:
Tell travis to test go1.9beta2, fix the build break in gotestmain, and enable the new parallel compilation support in the go compiler.

### DIFF
--- a/microfactory/microfactory.go
+++ b/microfactory/microfactory.go
@@ -69,6 +69,7 @@ var (
 
 	goToolDir = filepath.Join(runtime.GOROOT(), "pkg", "tool", runtime.GOOS+"_"+runtime.GOARCH)
 	goVersion = findGoVersion()
+	isGo18    = strings.Contains(goVersion, "go1.8")
 )
 
 func findGoVersion() string {
@@ -276,6 +277,9 @@ func (p *GoPackage) Compile(outDir, trimPath string) error {
 		"-o", p.output,
 		"-p", p.Name,
 		"-complete", "-pack", "-nolocalimports")
+	if !isGo18 {
+		cmd.Args = append(cmd.Args, "-c", fmt.Sprintf("%d", runtime.NumCPU()))
+	}
 	if race {
 		cmd.Args = append(cmd.Args, "-race")
 		fmt.Fprintln(hash, "-race")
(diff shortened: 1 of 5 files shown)


microfactory/microfactory.go
```diff
@@ -69,6 +69,7 @@ var (
 
 	goToolDir = filepath.Join(runtime.GOROOT(), "pkg", "tool", runtime.GOOS+"_"+runtime.GOARCH)
 	goVersion = findGoVersion()
+	isGo18    = strings.Contains(goVersion, "go1.8")
 )
 
 func findGoVersion() string {
```
colincross: This function seems inaccurate.  If you can assume that microfactory was built with the go version in GOROOT, you can use runtime.Version(), but for the beta won't all 3 of these methods return a SHA and not a string?

microfactory/microfactory.go
```diff
@@ -69,6 +69,7 @@ var (
 
 	goToolDir = filepath.Join(runtime.GOROOT(), "pkg", "tool", runtime.GOOS+"_"+runtime.GOARCH)
 	goVersion = findGoVersion()
+	isGo18    = strings.Contains(goVersion, "go1.8")
```
colincross: A possible hack that may be more accurate, check if go/build.Default.ReleaseTags contains "go1.8"?

microfactory/microfactory.go
danw: I found out last time that you can't, since microfactory may be rebuilding itself after a Go upgrade, so we really want the go version that's in the GOROOT, not the one that previously existed and compiled microfactory.

In this case it's not that big of an issue, since you'd just lose some perf. The other use case of ensuring that the binaries got rebuilt when the go version changed was a bigger issue.

For the rebuilding case, any unique string works. For this -c case, the only time it would fail is if you used a development version that didn't have -c support (so a master/beta build before -c was added), since we're just blacklisting go1.8 (and require os.Executable so <1.8 isn't supported)

microfactory/microfactory.go
danw: We'd need to check to see if ReleaseTags does not include go1.9, since it'll include all previous version tags. I'll try that out.

danw:
Ok, updated the last workaround commit. Turned out to be a gimme bug instead of a go1.9rc1 bug (go1.9rc1 is just able to get further than go1.9beta2 would have been able to get without a GOROOT)

PR #166, opened 2017-07-28
Add TopDownMutatorContext.CreateModule

colincross:
Allow a module to create other modules as if they were specified
in a blueprint file.  CreateModule takes the name of a module type,
calls the factory function the module type, and then inserts
properties into the property structs returned by the factory.

### DIFF
--- a/module_ctx.go
+++ b/module_ctx.go
@@ -138,6 +138,7 @@ type BaseModuleContext interface {
 	GlobWithDeps(pattern string, excludes []string) ([]string, error)
 
 	Fs() pathtools.FileSystem
+	AddNinjaFileDeps(deps ...string)
 
 	moduleInfo() *moduleInfo
 	error(err error)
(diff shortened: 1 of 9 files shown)


module_ctx.go
```diff
@@ -737,6 +741,25 @@ func (mctx *mutatorContext) Rename(name string) {
 	mctx.rename = append(mctx.rename, rename{mctx.module.group, name})
 }
 
+// Create a new module by calling the factory method for the specified moduleType.  The module
+// will be empty, as if it were created by a blueprint file definition that contained no properties.
+// The caller should use AppendProperties
```
danw: This comment needs updating -- properties are passed in, not appended with AppendProperties

module_ctx.go
```diff
@@ -737,6 +741,25 @@ func (mctx *mutatorContext) Rename(name string) {
...
+func (mctx *mutatorContext) CreateModule(moduleType string, props ...interface{}) {
+	module := mctx.context.newModule(moduleType)
+	if module == nil {
+		panic(fmt.Errorf("Unknown module type %q", moduleType))
```
danw: Should we take into account ignoreUnknownModuleTypes? I'm leaning towards no, but it could be useful.

colincross:
Switch to append semantics for blueprint unpacking and CreateModule.

colincross:
s/Switch/Switched/

PR #158, opened 2017-06-28
Remove src.test after running Travis tests

mathjeff:
This should prevent the running of Travis tests from causing
a filesystem loop src.test/test_tree/blueprint -> ../..

(except if the test unexpectedly exits)

Bug: 63100441
Test: NINJA=... ./tests/test_tree_tests.sh; stat src.test 2>/dev/null \
      && echo "src.test should not exist"

Change-Id: Ie01fe8999c738dca166e3fabde9d9391f6b70210

PR #168, opened 2017-08-11
File list

mathjeff:
Support for a custom list of Blueprints files to parse

### DIFF
--- a/bootstrap.bash
+++ b/bootstrap.bash
@@ -62,9 +62,6 @@ fi
 # Go toolchain properties.
 [ -z "$GOROOT" ] && GOROOT=`go env GOROOT`
 
-# If RUN_TESTS is set, behave like -t was passed in as an option.
-[ ! -z "$RUN_TESTS" ] && EXTRA_ARGS="-t"
-
 usage() {
     echo "Usage of ${BOOTSTRAP}:"
     echo "  -h: print a help message and exit"
@@ -76,7 +73,7 @@ usage() {
 while getopts ":b:ht" opt; do
     case $opt in
         b) BUILDDIR="$OPTARG";;
-        t) EXTRA_ARGS="-t";;
+        t) RUN_TESTS=true;;
         h)
             usage
             exit 1
@@ -93,6 +90,15 @@ while getopts ":b:ht" opt; do
     esac
 done
 
+# If RUN_TESTS is set, behave like -t was passed in as an option.
+[ ! -z "$RUN_TESTS" ] && EXTRA_ARGS="${EXTRA_ARGS} -t"
+
+# Allow the caller to pass in a list of module files
+if [ -z "${BLUEPRINT_LIST_FILE}" ]; then
+  BLUEPRINT_LIST_FILE="${BUILDDIR}/.bootstrap/bplist"
+fi
+EXTRA_ARGS="${EXTRA_ARGS} -l ${BLUEPRINT_LIST_FILE}"
+
 mkdir -p $BUILDDIR/.minibootstrap
 
 echo "bootstrapBuildDir = $BUILDDIR" > $BUILDDIR/.minibootstrap/build.ninja
@@ -101,11 +107,14 @@ echo "extraArgs = $EXTRA_ARGS" >> $BUILDDIR/.minibootstrap/build.ninja
 echo "builddir = $NINJA_BUILDDIR" >> $BUILDDIR/.minibootstrap/build.ninja
 echo "include $BLUEPRINTDIR/bootstrap/build.ninja" >> $BUILDDIR/.minibootstrap/build.ninja
 
-echo "BLUEPRINT_BOOTSTRAP_VERSION=1" > $BUILDDIR/.blueprint.bootstrap
+echo "BLUEPRINT_BOOTSTRAP_VERSION=2" > $BUILDDIR/.blueprint.bootstrap
 echo "SRCDIR=\"${SRCDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "BLUEPRINTDIR=\"${BLUEPRINTDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "NINJA_BUILDDIR=\"${NINJA_BUILDDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "GOROOT=\"${GOROOT}\"" >> $BUILDDIR/.blueprint.bootstrap
+echo "TOPNAME=\"${TOPNAME}\"" >> $BUILDDIR/.blueprint.bootstrap
+
+touch "${BUILDDIR}/.out-dir"
 
 if [ ! -z "$WRAPPER" ]; then
     cp $WRAPPER $BUILDDIR/
--- a/bootstrap/command.go
+++ b/bootstrap/command.go
@@ -31,14 +31,15 @@ import (
 )
 
 var (
-	outFile    string
-	depFile    string
-	docFile    string
-	cpuprofile string
-	memprofile string
-	traceFile  string
-	runGoTests bool
-	noGC       bool
+	outFile        string
+	depFile        string
+	docFile        string
+	cpuprofile     string
+	memprofile     string
+	traceFile      string
+	runGoTests     bool
+	noGC           bool
+	moduleListFile string
 
 	BuildDir      string
 	NinjaBuildDir string
@@ -56,6 +57,7 @@ func init() {
 	flag.StringVar(&memprofile, "memprofile", "", "write memory profile to file")
 	flag.BoolVar(&noGC, "nogc", false, "turn off GC for debugging")
 	flag.BoolVar(&runGoTests, "t", false, "build and run go tests during bootstrap")
+	flag.StringVar(&moduleListFile, "l", "", "file that lists filepaths to parse")
 }
 
 func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...string) {
@@ -93,11 +95,21 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 		fatalf("no Blueprints file specified")
 	}
 
+	SrcDir = filepath.Dir(flag.Arg(0))
+	if moduleListFile != "" {
+		ctx.SetModuleListFile(moduleListFile)
+		extraNinjaFileDeps = append(extraNinjaFileDeps, moduleListFile)
+	} else {
+		fatalf("-l <moduleListFile> is required and must be nonempty")
+	}
+	filesToParse, err := ctx.ListModulePaths(SrcDir)
+	if err != nil {
+		fatalf("could not enumerate files: %v\n", err.Error())
+	}
+
 	if NinjaBuildDir == "" {
 		NinjaBuildDir = BuildDir
 	}
-
-	SrcDir = filepath.Dir(flag.Arg(0))
 
 	stage := StageMain
 	if c, ok := config.(ConfigInterface); ok {
@@ -110,6 +122,7 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 		stage: stage,
 		topLevelBlueprintsFile: flag.Arg(0),
 		runGoTests:             runGoTests,
+		moduleListFile:         moduleListFile,
 	}
 
 	ctx.RegisterBottomUpMutator("bootstrap_plugin_deps", pluginDeps)
@@ -120,7 +133,7 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 
 	ctx.RegisterSingletonType("glob", globSingletonFactory(ctx))
 
-	deps, errs := ctx.ParseBlueprintsFiles(bootstrapConfig.topLevelBlueprintsFile)
+	deps, errs := ctx.ParseFileList(filepath.Dir(bootstrapConfig.topLevelBlueprintsFile), filesToParse)
 	if len(errs) > 0 {
 		fatalErrors(errs)
 	}
@@ -149,7 +162,7 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 	deps = append(deps, extraDeps...)
 
 	buf := bytes.NewBuffer(nil)
-	err := ctx.WriteBuildFile(buf)
+	err = ctx.WriteBuildFile(buf)
 	if err != nil {
 		fatalf("error generating Ninja file contents: %s", err)
 	}
--- a/context.go
+++ b/context.go
@@ -878,16 +878,6 @@ func (c *Context) parseOne(rootDir, filename string, reader io.Reader,
 	}
 	file.Name = relBlueprintsFile
 
-	subdirs, subdirsPos, err := getLocalStringListFromScope(scope, "subdirs")
-	if err != nil {
-		errs = append(errs, err)
-	}
-
-	optionalSubdirs, optionalSubdirsPos, err := getLocalStringListFromScope(scope, "optional_subdirs")
-	if err != nil {
-		errs = append(errs, err)
-	}
-
 	build, buildPos, err := getLocalStringListFromScope(scope, "build")
 	if err != nil {
 		errs = append(errs, err)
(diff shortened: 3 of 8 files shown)


bootstrap/command.go
```diff
@@ -89,8 +110,13 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
...
 
-	if flag.NArg() != 1 {
-		fatalf("no Blueprints file specified")
+	filesToParse, err := listFiles()
+	if err != nil {
+		if flag.NArg() != 1 {
+			fatalf("no Blueprints file specified, and "+
+				"could not read list file because: %v\n", err.Error())
+		}
+		filesToParse = []string{flag.Arg(0)}
```
mathjeff: I suspect that we don't want to be passing both a list of files via listFiles() and also a root file via flag.Arg(0) . I'm still investigating a better way make that work

colincross:
These patches are on top of the Android history, can you rebase them on top of upstream master?

context.go
```diff
@@ -689,9 +690,9 @@ loop:
 	return deps, errs
 }
 
-type FileHandler func(*parser.File)
+type FileVisitor func(*parser.File)
```
colincross: Handler makes more sense than Visitor to me, Visitor implies walking something that is already parsed to me.  The implementation above is also still called handleOneFile.

context.go
```diff
@@ -514,17 +514,18 @@ func (c *Context) SetAllowMissingDependencies(allowMissingDependencies bool) {
 	c.allowMissingDependencies = allowMissingDependencies
 }
 
-// Parse parses a single Blueprints file from r, creating Module objects for
-// each of the module definitions encountered.  If the Blueprints file contains
-// an assignment to the "subdirs" variable, then the subdirectories listed are
-// searched for Blueprints files returned in the subBlueprints return value.
-// If the Blueprints file contains an assignment to the "build" variable, then
-// the file listed are returned in the subBlueprints return value.
+// parseOne parses a single Blueprints file from <reader>, creating Module
```
colincross: godoc style doesn't use <> around parameters.

context.go
```diff
@@ -783,11 +785,15 @@ func (c *Context) MockFileSystem(files map[string][]byte) {
...
-func (c *Context) parseBlueprintsFile(filename string, scope *parser.Scope, rootDir string,
+// parseOneCh parses a single Blueprints file, and sends results through the provided channels
+//
+// Errors are returned through errsCh.
+// Any defined modules are returned through modulesCh.
+// Any sub-Blueprints files are returned through blueprintsCh.
+// Any dependencies on Blueprints files or directories are returned through depsCh.
+// When processing completes (and all other channel responses have been sent successfully),
+// an empty struct is passed into doneCh.
+func (c *Context) parseOneCh(filename string, scope *parser.Scope, rootDir string,
```
colincross: It's not parsing a channel, its parsing a blueprints file

bootstrap/command.go
```diff
@@ -58,6 +61,27 @@ func init() {
 	flag.BoolVar(&runGoTests, "t", false, "build and run go tests during bootstrap")
 }
 
+func listFiles(srcDir string) (filePaths []string, err error) {
+	listFile := ModulesFile
```
colincross: Take listFile as a parameter so that only Main is referencing the global variables

bootstrap/command.go
colincross: Yeah, we might delete the root Android.bp file.  On the other hand, people will expect it to exist so maybe we will replace it with one containing a comment that explains the new system, in which case we can punt on supporting invoking blueprint without a root file until later.

context.go
```diff
@@ -641,15 +652,33 @@ func (c *Context) WalkBlueprintsFiles(rootFile string, visitor FileVisitor) (dep
 			c.parseOneCh(blueprint.string, blueprint.Scope, rootDir,
 				errsCh, fileCh, blueprintsCh, depsCh)
 
+			descendants, hasDescendants := descendantsMap[blueprint.string]
+			if hasDescendants {
+				for _, descendant := range descendants {
+					blueprintsCh <- stringAndScope{descendant, parser.NewScope(blueprint.Scope)}
```
colincross: I think it makes more sense for the outer goroutine to handle the descendants.  I suggest removing blueprintCh, and returning the blueprint struct on doneCh.

context.go
```diff
@@ -1399,6 +1423,57 @@ func (c *Context) addInterVariantDependency(origModule *moduleInfo, tag Dependen
 	atomic.AddUint32(&c.depsModified, 1)
 }
 
+// parentFile returns the path of a file with the same name but in a parent directory
+// For example, if path = "a/b/Blueprints", then parent = "a/Blueprints"
+func parentFile(path string) (parent string, hasParent bool) {
```
colincross: Would it be easier to do the ancestor mapping on directories and ignore the file name?

context.go
```diff
@@ -1399,6 +1423,57 @@ func (c *Context) addInterVariantDependency(origModule *moduleInfo, tag Dependen
...
+		return "", false
+	}
+	return parentFile, true
+}
+
+// findBlueprintAncestors returns two maps linking parents to children
+// For example, if paths = []string{"a/b/c/Blueprints", "a/Blueprints"},
+// then ancestors = {"a/b/c/Blueprints":"a/Blueprints"}
+// and descendants = {"a/Blueprints", ["a/b/c/Blueprints"]}
+func findBlueprintAncestors(paths []string) (ancestors map[string]string, descendants map[string][]string) {
```
colincross: Returning descendants and a list of paths that have no ancestors would be more useful.  Or just return descendants, and put paths with no ancestors as descendants of "".

context.go
mathjeff: Does this mean to drop support for "subdirs="?

colincross: I guess leave blueprintCh, but also return blueprint on doneCh so the outer one can handle descendants?  Or leave it this way if you think its cleaner.

context.go
mathjeff: thanks

context.go
mathjeff: done

context.go
mathjeff: done

bootstrap/command.go
mathjeff: thanks

context.go
mathjeff: ok

context.go
mathjeff: done

context.go
```diff
@@ -642,15 +733,31 @@ func (c *Context) WalkBlueprintsFiles(rootFile string, visitor FileVisitor) (dep
...
+		if hasDescendants {
+			for _, descendant := range descendants {
+				foundParseableBlueprint(stringAndScope{descendant, parser.NewScope(blueprint.Scope)})
+			}
+		}
+	}
 
-	var pending []stringAndScope
+	// begin parsing any files that have no ancestors
+	for _, path := range descendantsMap[""] {
```
colincross: I'm not sure if its better, but this could be startParseDescendants(stringAndScope{"", parser.NewScope(nil)})

context.go
```diff
@@ -669,18 +776,18 @@ loop:
...
-				continue
-			}
-			startParseBlueprintsFile(blueprint)
-		case <-doneCh:
+			foundParseableBlueprint(blueprint)
+		case blueprint := <-doneCh:
 			activeCount--
-			if len(pending) > 0 {
-				startParseBlueprintsFile(pending[len(pending)-1])
+			deps = append(deps, blueprint.string)
```
colincross: Can this go in startParseDescendants instead?

mathjeff:
Hmm it looks like the latest version of this patchset results in slower execution than without it, so we'll want to wait to actually merge it until that's fixed

context.go
```diff
@@ -792,51 +906,21 @@ func (c *Context) parseOne(rootDir, filename string, reader io.Reader,
 	}
 	file.Name = relBlueprintsFile
 
-	subdirs, subdirsPos, err := getLocalStringListFromScope(scope, "subdirs")
```
colincross: Are these deletions the only thing required to remove subdirs support?  If so, lets leave them in and then we can figure out what we want to do with legacy subdirs support in a separate patch.  Using both file list and subdirs together shouldn't cause a problem because startParseBlueprintsFile checks if the file has already been parsed.

context.go
```diff
@@ -1417,6 +1465,43 @@ func (c *Context) addInterVariantDependency(origModule *moduleInfo, tag Dependen
...
+		if alreadyFound {
+			return nil, fmt.Errorf("Found two Blueprint files in directory %v : %v and %v", dir, filesByDir[dir], path)
+		}
+		filesByDir[dir] = path
+	}
+	// generate the descendants map
+	descendants = make(map[string][]string, len(filesByDir))
+	for childDir, childFile := range filesByDir {
+		prevAncestorDir := childDir
+		for {
```
colincross: Not sure if its better, but this could be:
for ancestorDir := filepath.Dir(prevAncestorDir); ancestorDir != prevAncestorDir {

context.go
```diff
@@ -1417,6 +1465,43 @@ func (c *Context) addInterVariantDependency(origModule *moduleInfo, tag Dependen
...
+			ancestorDir := filepath.Clean(filepath.Dir(prevAncestorDir))
```
colincross: Dir will always return a clean path, no need to Clean it again.

context.go
mathjeff: ok

context.go
mathjeff: done

context.go
mathjeff: done

context.go
mathjeff: thanks

context.go
mathjeff: It does also need to add childFile as a descendant of descendants[""] if it reaches the root directory. Could you clarify on the syntax that you would like, or shall we just leave it?

bootstrap/command.go
```diff
@@ -148,7 +148,7 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 	deps = append(deps, extraDeps...)
```
mathjeff: I suspect a dependency somewhere, possibly this <deps> , needs to include the $BLUEPRINT_LIST_FILE in order to rerun the right invocations of Blueprint when a new Android.bp is created.

context.go
mathjeff: Hmm it turns out that there exists a symlink (not in a .git repo, but specified by a manifest) named Android.bp pointing to an Android.bp in a subdirectory, which has a nonempty value for "subdirs".

Previously the parent Android.bp was the one that was getting parsed by Blueprint, so the relative file path listed in the child is supposed to be interpreted relative to the parent. Then when this change causes Blueprint to search for Android.bp files, it finds the child Android.bp and looks for the target of the subdirs relative to the child, and fails to open the result.

Do you think that that makes it worthwhile to remove subdirs support at the same time, or whether we want to fix the Android.bp files (probably involving renaming them and updating some manifests to repoint the symlink) ?

bootstrap/command.go
mathjeff: Yeah that ended up being added in the next patchset

mathjeff:
It looks like the Travis build failed because tests/test_tree_tests.sh uses a symlink: blueprint -> ../.. 

It seems to me that the most convenient way to fix it starts with merging https://github.com/google/blueprint/pull/158

colincross:
Merged it an reran travis but it still fails

mathjeff:
Ping!

Does this patch need more review or is it only blocked on merging the removals of various non-working projects?

bootstrap/command.go
```diff
@@ -89,16 +91,19 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
 		defer trace.Stop()
 	}
 
-	if flag.NArg() != 1 {
-		fatalf("no Blueprints file specified")
```
danw: Why are you removing the NArg check?

bootstrap.bash
```diff
@@ -101,11 +106,14 @@ echo "extraArgs = $EXTRA_ARGS" >> $BUILDDIR/.minibootstrap/build.ninja
 echo "builddir = $NINJA_BUILDDIR" >> $BUILDDIR/.minibootstrap/build.ninja
 echo "include $BLUEPRINTDIR/bootstrap/build.ninja" >> $BUILDDIR/.minibootstrap/build.ninja
 
-echo "BLUEPRINT_BOOTSTRAP_VERSION=1" > $BUILDDIR/.blueprint.bootstrap
+echo "BLUEPRINT_BOOTSTRAP_VERSION=2" > $BUILDDIR/.blueprint.bootstrap
 echo "SRCDIR=\"${SRCDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "BLUEPRINTDIR=\"${BLUEPRINTDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "NINJA_BUILDDIR=\"${NINJA_BUILDDIR}\"" >> $BUILDDIR/.blueprint.bootstrap
 echo "GOROOT=\"${GOROOT}\"" >> $BUILDDIR/.blueprint.bootstrap
+echo "TOPNAME=\"${TOPNAME}\"" >> $BUILDDIR/.blueprint.bootstrap
```
danw: Why does TOPNAME need to be preserved?

bootstrap/command.go
```diff
@@ -161,9 +168,19 @@ func Main(ctx *blueprint.Context, config interface{}, extraNinjaFileDeps ...stri
...
-		if err != nil {
-			fatalf("error writing depfile: %s", err)
+		if moduleListFile != "" {
+			err := deptools.WriteDepFile(depFile, outFile, deps)
+			if err != nil {
+				fatalf("error writing depfile: %s", err)
+			}
+		} else {
+			// If no file list is provided, then essentially every directory is a dependency
+			// Rather than emitting a list of every directory, we just skip making the deps file.
```
danw: The depfile contains more than just the Blueprints/Android.bp files, I don't think we can remove it entirely. Or is this just triggering it to just always build?

context.go
```diff
@@ -106,7 +107,9 @@ type Context struct {
 	globs    map[string]GlobPath
 	globLock sync.Mutex
 
-	fs pathtools.FileSystem
+	fs                 pathtools.FileSystem
+	sourceFinder       func(baseDir string) ([]string, error)
+	ExtraNinjaFileDeps []string
```
danw: Why not just returns a list of deps from SetModuleListFile like all of the other ctx functions?

context.go
```diff
@@ -285,6 +289,36 @@ func NewContext() *Context {
 	return ctx
 }
 
+// searches for Blueprints files under baseDir
+func findBlueprintsFromDir(baseDir string) (filePaths []string, err error) {
```
danw: If we're going to implement this here, can we also export a list of dependencies so that the dep file actually works?

tests/test_tree_tests.sh
```diff
@@ -25,39 +26,17 @@ cp -r "${TOP}" "${SRCDIR}/blueprint"
...
 OLDTIME_BOOTSTRAP=$(mtime .bootstrap/build.ninja)
 OLDTIME=$(mtime build.ninja)
 
-sleep 2
-./blueprint.bash
-
-if [ ${OLDTIME} != $(mtime build.ninja) ]; then
-    echo "unnecessary build.ninja regeneration for null build" >&2
-    exit 1
-fi
```
danw: Why are these tests being removed?

bootstrap/command.go
mathjeff: Hmm, that must've been a mistake

bootstrap/command.go
mathjeff: This was triggering it to always rebuild, although the nextpatchset takes your suggestion of always generating the listfile, so this code gets removed

context.go
mathjeff: thanks

context.go
mathjeff: this is removed in the next patchset at your suggestion

bootstrap.bash
mathjeff: Oops, it wasn't needed for this patchset. It is needed for the next patchset after re-adding the bash search though.

tests/test_tree_tests.sh
mathjeff: they didn't work in this patchset, but the nextpatchset fixes them and keeps them in

PR #169, opened 2017-08-23
Make microfactory support go build tags

danw:
This is necessary to support other go code that uses build tags to control which files build on which platforms. For example, the github.com/golang/protobuf/proto package includes two implementations of pointer:

pointer_reflect:
```go
// ...

// +build appengine js

// This file contains an implementation of proto field accesses using package reflect.
// It is slower than the code in pointer_unsafe.go but it avoids package unsafe and can
// be used on App Engine.
```

pointer_unsafe.go:
```go
// ...

// +build !appengine,!js

// This file contains the implementation of the proto field accesses using package unsafe.
```

Without this change, microfactory tries to include both files, which breaks, since that enables multiple implementations of the same symbols. This implements the logic to find the +build comments (following the spirit of the functionality in go/build, even though they make different
choices of parsers), and the logic to match the tags to the current system (which isn't exported as public API, so we have a simplified version).

Since we're only parsing the comments before the import statement, this does not significantly affect performance.

Test: Build github.com/golang/protobuf/proto

### DIFF
--- a/microfactory/microfactory.go
+++ b/microfactory/microfactory.go
@@ -47,6 +47,7 @@ import (
 	"flag"
 	"fmt"
 	"go/ast"
+	"go/build"
 	"go/parser"
 	"go/token"
 	"io"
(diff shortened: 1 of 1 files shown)


microfactory/microfactory.go
```diff
@@ -236,6 +293,18 @@ func (p *GoPackage) findDeps(config *Config, path string, allPackages *linkedDep
 	localDeps := make(map[string]bool)
 
 	for filename, astFile := range foundPkg.Files {
+		ignore := false
+		for _, commentGroup := range astFile.Comments {
+			for _, comment := range commentGroup.List {
+				if matches, ok := parseBuildComment(comment.Text); ok && !matches {
+					ignore = true
```
mathjeff: Do we want to break from the loop too?

microfactory/microfactory.go
```diff
@@ -199,6 +200,62 @@ func (p *GoPackage) FindDeps(config *Config, path string) error {
...
+
+	return false
+}
+
+func parseBuildComment(comment string) (matches, ok bool) {
+	if !strings.HasPrefix(comment, "//") {
+		return false, false
+	}
+	for i, c := range comment {
+		if i < 2 || c == ' ' || c == '\t' {
```
mathjeff: I suspect you can do "range comment[2:]"

microfactory/microfactory.go
```diff
@@ -217,7 +274,7 @@ func (p *GoPackage) findDeps(config *Config, path string, allPackages *linkedDep
 			return false
 		}
 		return true
-	}, parser.ImportsOnly)
+	}, parser.ImportsOnly|parser.ParseComments)
```
mathjeff: Not that I'm really that concerned about performance, but which part is it that implements "we're only parsing the comments before the import statement"?

microfactory/microfactory.go
```diff
@@ -199,6 +200,62 @@ func (p *GoPackage) FindDeps(config *Config, path string) error {
...
+	}
+	for _, tag := range build.Default.ReleaseTags {
+		if tag == name {
+			return true
+		}
+	}
+
+	return false
+}
+
```
mathjeff: How would you feel about a comment like "tells whether the given comment requests that the current file be included/excluded from compilation"?

PR #170, opened 2017-08-24
Clearer error in case of Android.bp being unreadable

mathjeff:
Bug: 64600838
Test: mkdir errtest \
      && ln -s /tmp/dontexist errtest/Android.bp \
      # and add errtest to ./Android.bp \
      && m nothing \
      # and check that the error message mentions a symlink

Change-Id: I841ec12d613f61ccc3396538062bee48c8c1ca27

### DIFF
--- a/context.go
+++ b/context.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -793,6 +794,14 @@ func (c *Context) parseBlueprintsFile(filename string, scope *parser.Scope, root
 
 	f, err := c.fs.Open(filename)
 	if err != nil {
+		stats, statErr := c.fs.Lstat(filename)
+		if statErr == nil {
+			if stats.Mode()&os.ModeSymlink != 0 {
+				err = fmt.Errorf("%v is a symlink that could not be opened, perhaps its destination doesn't exist: %v", filename, err)
```
colincross: Can you use readlink to find the destination of the link, stat that, and print a even clearer error if it doesn't exist?

mathjeff: Sure. Am I correct to assume that a single ReadLink is all we'd like to do (rather than a recursive ReadLink for a potentially maximally informative error message in the case of a chain of symlinks whose final destination doesn't exist)?

colincross: Good point, but to handle chained readlinks you would need to deal with loops, which sounds like more trouble than it is worth.

mathjeff:
Hmm, GitHub says that this is approved but it tells me "Only those with write access to this repository can merge pull requests."

Do you think it makes more sense for you to merge it or for me to have write access?

PR #173, opened 2017-09-28
Some clarifications in preparation to automatically order linker depe…

mathjeff:
…ndencies

Test: Browse the code and determine whether it's easier to understand
Bug: 66260943
Change-Id: I9ee8c1d5461008b489b052f23ca78096d0089fd8

### DIFF
--- a/context.go
+++ b/context.go
@@ -42,7 +42,7 @@ const maxErrors = 10
 
 // A Context contains all the state needed to parse a set of Blueprints files
 // and generate a Ninja file.  The process of generating a Ninja file proceeds
-// through a series of four phases.  Each phase corresponds with a some methods
+// through a series of four phases.  Each phase corresponds with some methods
 // on the Context object
 //
 //         Phase                            Methods
@@ -229,9 +229,9 @@ func (vm variationMap) clone() variationMap {
 	return newVm
 }
 
-// Compare this variationMap to another one.  Returns true if the every entry in this map
-// is either the same in the other map or doesn't exist in the other map.
-func (vm variationMap) subset(other variationMap) bool {
+// Compare this variationMap to another one.  Returns true if every key that is
+// shared by both maps also has the same value in each map
+func (vm variationMap) intersectionMatches(other variationMap) bool {
 	for k, v1 := range vm {
 		if v2, ok := other[k]; ok && v1 != v2 {
 			return false
@@ -960,7 +960,7 @@ func getStringFromScope(scope *parser.Scope, v string) (string, scanner.Position
 }
 
 // Clones a build logic module by calling the factory method for its module type, and then cloning
-// property values.  Any values stored in the module object that are not stored in properties
+// property values.  Any values stored in the module object that are not stored in Properties
 // structs will be lost.
 func (c *Context) cloneLogicModule(origModule *moduleInfo) (Module, []interface{}) {
 	newLogicModule, newProperties := origModule.factory()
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -1227,6 +1227,7 @@ func (c *Context) findMatchingVariant(module *moduleInfo, possible []*moduleInfo
 }
 
 func (c *Context) addDependency(module *moduleInfo, tag DependencyTag, depName string) []error {
+
```
colincross: I usually only insert an empty line at the top of the function if the function signature wraps to multiple lines, since gofmt will indent the wrapped signature the same as the first statement.

context.go
```diff
@@ -3208,7 +3208,7 @@ func (c *Context) writeLocalBuildActions(nw *ninjaWriter,
 	return nil
 }
 
-func beforeInModuleList(a, b *moduleInfo, list []*moduleInfo) bool {
+func aPrecedesBInModuleList(a, b *moduleInfo, list []*moduleInfo) bool {
```
colincross: I agree the original is hard to parse, but I don't think this helps because any callers don't have an "a" and a "b", those are local to the function.  Maybe just add godoc that describes that this checks if the first argument is before the second argument in the list?

module_ctx.go
```diff
@@ -278,6 +282,7 @@ type moduleContext struct {
 	handledMissingDeps bool
 }
 
+// TODO: expain how this differs from logicModule.Name() : does this relativize/contextualize paths?
```
colincross: See the comment on Module.Name().  logicModule.Name() is called exactly once, and returns what the module is initially know as.  moduleInfo.Name() returns the current name, which is initially the value of logicModule.Name(), but could be renamed by MutatorContext.Rename().

context.go
```diff
@@ -960,7 +960,7 @@ func getStringFromScope(scope *parser.Scope, v string) (string, scanner.Position
 }
 
 // Clones a build logic module by calling the factory method for its module type, and then cloning
-// property values.  Any values stored in the module object that are not stored in properties
+// property values.  Any values stored in the module object that are not stored in Properties
```
colincross: I don't think Properties should be capitalized, that makes me think it is a Go type, but we are using it to refer to a class of structs.

module_ctx.go
```diff
@@ -560,17 +565,17 @@ type BottomUpMutatorContext interface {
 // after parsing all Blueprint files, but before generating any build rules,
 // and is always called on dependencies before being called on the depending module.
 //
-// The Mutator function should only modify members of properties structs, and not
+// The Mutator function should only modify members of Properties structs, and not
```
colincross: I don't think this should be capitalized

module_ctx.go
```diff
@@ -679,10 +684,10 @@ func (mctx *mutatorContext) AddReverseDependency(module Module, tag DependencyTa
 	})
 }
 
-// AddVariationDependencies adds deps as dependencies of the current module, but uses the variations
-// argument to select which variant of the dependency to use.  A variant of the dependency must
-// exist that matches the all of the non-local variations of the current module, plus the variations
-// argument.
+// AddVariationDependencies adds deps as dependencies of the current module, but uses the
+// variations argument to select which variant of the dependency to use.  The dependency must
+// provide a a variant whose Variation coordinates match each non-local Variation coordinate of the current module and each coordinate
```
colincross: I don't understand the use of the word coordinates here.

package_ctx.go
```diff
@@ -39,9 +39,13 @@ import (
 //     var (
 //         pctx = NewPackageContext("path/to/blah")
 //
+//         // Should appear in the generated ninja file as "g.path.to.blah.myPrivateVar = abcdef"
 //         myPrivateVar = pctx.StaticVariable("myPrivateVar", "abcdef")
+//
+//         // Should appear in the generated ninja file as "g.path.to.blah.MyExportedVar = ${myPrivateVar} 123456!"
```
colincross: ${g.path.to.blah.myPrivateVar}

package_ctx.go
```diff
@@ -39,9 +39,13 @@ import (
 //     var (
 //         pctx = NewPackageContext("path/to/blah")
 //
+//         // Should appear in the generated ninja file as "g.path.to.blah.myPrivateVar = abcdef"
```
colincross: Maybe not worth mentioning here, but live_tracker.go strips the variables and rules down to ones that are referenced through a build statement.

mathjeff:
never figured out a wording that we all liked (similar to "coordinates") to make extra obvious the distinction between "variation" and "variant"

PR #174, opened 2017-10-03
Fix recursive glob on MockFs

colincross:
Recursive globs on MockFs were ending up in an unmocked os.Lstat
call in walkAllDirs.  Move walkAllDirs into the FileSystem interface
as ListDirsRecursive.

Also duplicate the glob tests on both the real filesystem and on
MockFs.

Test: glob_test.go
Change-Id: Ia6b6b5eecdd17955a49d684a0fd5e55df05cfe62

### DIFF
--- a/pathtools/fs.go
+++ b/pathtools/fs.go
@@ -22,6 +22,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
+	"strings"
 )
 
 // Based on Andrew Gerrand's "10 things you (probably) dont' know about Go"
@@ -53,6 +55,8 @@ func MockFs(files map[string][]byte) FileSystem {
 		fs.all = append(fs.all, d)
 	}
 
+	sort.Strings(fs.all)
+
 	return fs
 }
 
@@ -63,6 +67,7 @@ type FileSystem interface {
 	glob(pattern string) (matches []string, err error)
 	IsDir(name string) (bool, error)
 	Lstat(name string) (os.FileInfo, error)
+	ListDirsRecursive(name string) (dirs []string, err error)
 }
 
 // osFs implements FileSystem using the local disk.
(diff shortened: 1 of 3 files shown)


pathtools/fs.go
```diff
@@ -160,3 +186,19 @@ func (m *mockFs) glob(pattern string) ([]string, error) {
 func (m *mockFs) Lstat(path string) (stats os.FileInfo, err error) {
 	return nil, errors.New("Lstat is not yet implemented in MockFs")
 }
+func (m *mockFs) ListDirsRecursive(name string) (dirs []string, err error) {
+	name = filepath.Clean(name)
+	dirs = append(dirs, name)
+	if name == "." {
+		name = ""
```
mathjeff: I guess it's worth noting that this assumes that no absolute paths were entered into m.all. It seems to me that it'd be nice to either add an assertion confirming that whenever a path is added to m.all that it isn't absolute, or to update this to handle that case.

It looks like there's already an assertion in a test below - I'd just move it into the mockFs itself

pathtools/glob_test.go
```diff
@@ -441,37 +445,69 @@ var globTestCases = []struct {
...
+		"c/f/f.ext":  nil,
+		"c/g/g.ext":  nil,
+		"c/h/h":      nil,
+		"d.ext":      nil,
+		"e.ext":      nil,
+		".test/a":    nil,
+		".testing":   nil,
+		".test/.ing": nil,
+	})
+	for i, testCase := range globTestCases[16:17] {
```
colincross: Whoops, this was disabling most of the tests

pathtools/fs.go
colincross: I fixed the tests to test the absolute paths too.

colincross: Fixed again, MockFs globs will now only match an absolute path if given an absolute pattern.  The mock glob tests were actually finding the issue that you were describing, but I had incorrectly worked around it by using one MockFs that contained only relative paths and another that contained only absolute paths.

mathjeff:
Want to rebase onto github/master but otherwise looks good to me

PR #175, opened 2017-10-07
Have bpfmt show usage when run without any arguments

mathjeff:


bpfmt/bpfmt.go
```diff
@@ -40,17 +41,18 @@ func usage() {
...
-// If in == nil, the source is the contents of the file with the given filename.
-func processFile(filename string, in io.Reader, out io.Writer) error {
-	if in == nil {
-		f, err := os.Open(filename)
-		if err != nil {
-			return err
-		}
-		defer f.Close()
-		in = f
+func processFile(filename string, out io.Writer) error {
```
mathjeff: it turns out that this patch ("Refactor in preparation to detect when stdout is the input file") didn't end up being needed, but I kept it in anyway thinking that it was slightly cleaner

bpfmt/bpfmt.go
```diff
@@ -40,17 +41,18 @@ func usage() {
...
+	f, err := os.Open(filename)
+	if err != nil {
+		return err
 	}
+	defer f.Close()
+	in := f
```
colincross: This doesn't make sense any more, either rename f to in or pass f to processReader

bpfmt/bpfmt.go
```diff
@@ -36,8 +37,13 @@ func report(err error) {
 }
 
 func usage() {
+	usageViolation("")
+}
+
+func usageViolation(violation string) {
 	fmt.Fprintf(os.Stderr, "usage: bpfmt [flags] [path ...]\n")
 	flag.PrintDefaults()
+	fmt.Fprintln(os.Stderr, violation)
```
colincross: I think this traditionally is printed first before usage:

mathjeff: done

bpfmt/bpfmt.go
mathjeff: thanks

PR #176, opened 2017-10-17
Allow rules to specify order-only dependencies

colincross:
Commands that contain tools that don't affect the build results
may want an order-only dependency on the tool.  Allow rules
to specify order-only dependencies the same way they specify
implicit dependencies.

Test: builds
Change-Id: I3e0f886ae047b0fadf7a5c0dfeb9827d2c5c411d


danw:
If a command does not affect the build results, why not remove the use?

colincross:
soong_javac_wrapper.

PR #178, opened 2017-10-18
Add file locking to microfactory

danw:
To prevent against multiple instances attempting to rebuild the same executable, add a file lock to Build().

The trace file may not be consistent in this case, but everything that reads it should be able to handle a corrupted trace file.

Test: `while true; do <microfactory-enabled executable>; done` in parallel, and modify the code to microfactory (echo "//" >>microfactory.go)

### DIFF
--- a/microfactory/microfactory.go
+++ b/microfactory/microfactory.go
@@ -485,6 +485,18 @@ func Build(config *Config, out, pkg string) (*GoPackage, error) {
 		Name: "main",
 	}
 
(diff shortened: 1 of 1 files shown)


danw:
The trace file lock would be a bit trickier to get right -- we actually expect multiple writers to the trace file. When microfactory itself needs to be rebuilt, the trace is already partly written, and we execute the new version of microfactory where it'll append to the same trace file.

It's definitely possible, but I'm not convinced the complexity is worth it for the debugging feature in an edge case.

PR #181, opened 2017-10-31
Support parsing int64 number in Blueprint file

Nan-Zhang:
Support int64 number instead of int to be more fixed to num of bits so that the underlying arch won't affect overflow cases. Besides, refection: func (v Value) Int() int64 always cast to int64 no matter the input is int, int16, int32. Currently we always treat "-" as negative sign to bind to next value, and "+" as plus operator to add operands together.
So we allow:
a = 5 + -4 + 5 or a = -4 + 5
But we don't allow:
a = +5 + 4 + -4 since we don't treat "+" as a positive sign, otherwise,
a = 5 + +5 would exist which looks pretty weird. In the future, we may want fully support number calculator logic eg, "+"/"-" can be positive/negative sign or operator, and "(" and ")" will be considered to group expressions with a higher precedence.

int & uint properties within struct keeps unchanged, which is only allowed when tagged with 'blueprint:mutated'. We only allow ***int64** property instead of int64 property within struct since it does't make sense to do prepending or appending to int64.

### DIFF
--- a/parser/ast.go
+++ b/parser/ast.go
@@ -161,6 +161,7 @@ type Type int
 const (
 	BoolType Type = iota + 1
 	StringType
+	Int64Type
 	ListType
 	MapType
 )
@@ -171,6 +172,8 @@ func (t Type) String() string {
 		return "bool"
 	case StringType:
 		return "string"
+	case Int64Type:
+		return "int64"
 	case ListType:
 		return "list"
 	case MapType:
@@ -348,6 +351,31 @@ func (x *String) String() string {
 
 func (x *String) Type() Type {
 	return StringType
+}
+
+type Int64 struct {
+	LiteralPos scanner.Position
+	Value      int64
+}
+
+func (x *Int64) Pos() scanner.Position { return x.LiteralPos }
+func (x *Int64) End() scanner.Position { return x.LiteralPos }
+
+func (x *Int64) Copy() Expression {
+	ret := *x
+	return &ret
+}
+
+func (x *Int64) Eval() Expression {
+	return x
+}
+
+func (x *Int64) String() string {
+	return fmt.Sprintf("%q@%s", x.Value, x.LiteralPos)
+}
+
+func (x *Int64) Type() Type {
+	return Int64Type
 }
 
 type Bool struct {
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -114,7 +114,7 @@ func newParser(r io.Reader, scope *Scope) *parser {
 	p.scanner.Error = func(sc *scanner.Scanner, msg string) {
 		p.errorf(msg)
 	}
-	p.scanner.Mode = scanner.ScanIdents | scanner.ScanStrings |
+	p.scanner.Mode = scanner.ScanIdents | scanner.ScanInts | scanner.ScanStrings |
 		scanner.ScanRawStrings | scanner.ScanComments
 	p.next()
 	return p
@@ -330,6 +330,9 @@ func (p *parser) parseExpression() (value Expression) {
 	switch p.tok {
 	case '+':
 		return p.parseOperator(value)
+	case '-':
+		p.errorf("subtraction not supported: %s", p.scanner.String())
+		return value
 	default:
 		return value
 	}
@@ -355,6 +358,8 @@ func (p *parser) evaluateOperator(value1, value2 Expression, operator rune,
 			switch v := value.(type) {
 			case *String:
 				v.Value += e2.(*String).Value
+			case *Int64:
+				v.Value += e2.(*Int64).Value
 			case *List:
 				v.Values = append(v.Values, e2.(*List).Values...)
 			case *Map:
@@ -441,6 +446,8 @@ func (p *parser) parseValue() (value Expression) {
 	switch p.tok {
 	case scanner.Ident:
 		return p.parseVariable()
+	case '-', scanner.Int: // Integer might have '-' sign ahead ('+' is only treated as operator now)
+		return p.parseIntValue()
 	case scanner.String:
 		return p.parseStringValue()
 	case '[':
@@ -497,6 +504,32 @@ func (p *parser) parseStringValue() *String {
 		Value:      str,
 	}
 	p.accept(scanner.String)
+	return value
+}
+
+func (p *parser) parseIntValue() *Int64 {
+	var str string
+	literalPos := p.scanner.Position
+	if p.tok == '-' {
+		str += string(p.tok)
+		p.accept(p.tok)
+		if p.tok != scanner.Int {
+			p.errorf("expected int; found %s", scanner.TokenString(p.tok))
+			return nil
+		}
+	}
+	str += p.scanner.TokenText()
+	i, err := strconv.ParseInt(str, 10, 64)
+	if err != nil {
+		p.errorf("couldn't parse int: %s", err)
+		return nil
+	}
+
+	value := &Int64{
+		LiteralPos: literalPos,
+		Value:      i,
+	}
+	p.accept(scanner.Int)
 	return value
 }
 
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -110,6 +110,35 @@ var validParseTestCases = []struct {
 
 	{`
 		foo {
+			num: 4,
+		}
+		`,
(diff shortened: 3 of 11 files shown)


colincross:
I don't agree with most of the inlining done, and mixing the inlining with the functional changes makes it very hard to review.  Can you split that out to a separate change for discussion?

parser/parser.go
```diff
@@ -379,133 +388,88 @@ func (p *parser) evaluateOperator(value1, value2 Expression, operator rune,
...
-
-	return value
-
-}
-
 func (p *parser) parseValue() (value Expression) {
 	switch p.tok {
 	case scanner.Ident:
-		return p.parseVariable()
+		switch text := p.scanner.TokenText(); text {
```
mathjeff: My thoughts about whether to inline some functions since I know that a discussion about this is happening soon:

Either approach here seems reasonable to me. Since they're still clearly delineated by the case statements and likely to only be used once anyway I think it's ok to inline them, but as long as the function names are clear (which I think they are) I think also think it's easy to read them as separate functions.

parser/parser.go
```diff
@@ -379,133 +388,88 @@ func (p *parser) evaluateOperator(value1, value2 Expression, operator rune,
...
-		if p.eval {
-			if assignment, local := p.scope.Get(text); assignment == nil {
-				p.errorf("variable %q is not set", text)
-			} else {
-				if local {
-					assignment.Referenced = true
-				}
-				value = assignment.Value
+		p.accept(scanner.String)
+	case '-', scanner.Int: // Integer might have '-' sign ahead ('+' is only treated as operator currently.)
```
mathjeff: I'd replace "Integer might have '-' sign ahead" with "Integer might start with '-' " (or possibly "Integer might have preceding '-')

parser/parser.go
```diff
@@ -379,133 +388,88 @@ func (p *parser) evaluateOperator(value1, value2 Expression, operator rune,
...
+		var str string
+		literalPos := p.scanner.Position
+		if p.tok == '-' {
+			if p.scanner.Peek() < 48 || p.scanner.Peek() > 57 {
```
mathjeff: Could you replace "48" with " '0' " and " 57 " with " '9' " or otherwise make them into constants like perhaps numberZeroAsRune and numberNineAsRune?

parser/parser.go
```diff
@@ -533,20 +495,44 @@ func (p *parser) parseListValue() *List {
 	}
 }
 
-func (p *parser) parseMapValue() *Map {
-	lBracePos := p.scanner.Position
-	if !p.accept('{') {
-		return nil
+func (p *parser) parseMap(isModule bool) *Map {
```
mathjeff: If we do want the parsing of a map within a module to permit different separator semantics than the parsing of a map outside a module, then it seems to me that the variable should be named based on what it does as opposed to who currently sets it (since the question of who sets it could change in the future).

So if we do keep the "isModule" parameter, I'd rename "isModule" to "requireCurlyBrace" (and then negate its value to match the name).

Hmm on closer inspection the name "isModule" is actually unrelated to your change so if you don't feel like worrying about that then nevermind

parser/parser.go
```diff
@@ -355,14 +331,47 @@ func (p *parser) evaluateOperator(value1, value2 Expression, operator rune,
 			switch v := value.(type) {
 			case *String:
 				v.Value += e2.(*String).Value
+			case *Int64:
+				v.Value += e2.(*Int64).Value
```
mathjeff: It occurs to me that even though we currently support the "+" operator for lists, and we're now adding the ability to use integers, we don't necessarily have to support the "+" operator for integers too, which might be considered to be too much logic.

(And there's also the thought that even if we do disallow the "+" operator for lists, maybe that won't be enforced at the parser level but instead somewhere else)

Maybe we want to return an error and wait until there's a use case.

Maybe Dan or Colin will offer an opinion

mathjeff:
(I can't comment at this time unless I enter a non-blank top-level comment, so this comment is it)

Nan-Zhang:
after I did git reset --soft HEAD^, and then re-created a new commit, this pull request was automatically closed.. sorry about this.

Nan-Zhang:
it's back now. jeff pointed out that I just need re-open this pull-request, and it will be refreshed automatically with new commit.

Nan-Zhang:
I separate out the refactoring part from this adding int feature PR in case of bring more confusion. Once this get merged, I will bring that back.

parser/parser.go
```diff
@@ -330,6 +330,9 @@ func (p *parser) parseExpression() (value Expression) {
 	switch p.tok {
 	case '+':
 		return p.parseOperator(value)
+	case '-':
+		p.errorf("couldn't parse expression with '-' operator: %s", p.scanner.String())
```
mathjeff: I think we'd like to clarify a little why this isn't parseable, for example "using '-' to indicate subtraction is not permitted ('-' may only be used to indicate negative numbers)"

parser/ast.go
```diff
@@ -350,6 +353,31 @@ func (x *String) Type() Type {
 	return StringType
 }
 
+type Int64 struct {
+	LiteralPos scanner.Position
+	Value      int64
+}
+
+func (x *Int64) Pos() scanner.Position { return x.LiteralPos }
+func (x *Int64) End() scanner.Position { return x.LiteralPos }
```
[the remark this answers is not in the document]
Nan-Zhang: Dan/Colin can provide some opinions here. End() is not used for string, int and bool as I checked.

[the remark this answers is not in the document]
colincross: From https://godoc.org/github.com/google/blueprint/parser#Node:
    // End returns the position of the beginning of the last token in the Expression

I don't remember why I did it that way, it was definitely easier to calculate and I guess I didn't need the end of the last token for anything?  I think it is only used for ordering tokens.

parser/parser.go
colincross: Agree, maybe "subtraction not supported"?

parser/parser.go
colincross: I can't think of a case where + would be useful, but I also can't think of any other semantics we would want besides this and the implementation is trivial so I'm fine either way.

parser/parser.go
```diff
@@ -441,7 +446,9 @@ func (p *parser) parseValue() (value Expression) {
 	switch p.tok {
 	case scanner.Ident:
 		return p.parseVariable()
-	case scanner.String:
+	case '-', scanner.Int:
+		return p.parseIntValue()
+	case scanner.String: // Integer might have '-' sign ahead ('+' is only treated as operator now)
```
colincross: Why is the comment on this line?

parser/parser.go
```diff
@@ -500,6 +507,36 @@ func (p *parser) parseStringValue() *String {
 	return value
 }
 
+func (p *parser) parseIntValue() *Int64 {
+	var str string
+	literalPos := p.scanner.Position
+	if p.tok == '-' {
+		if p.scanner.Peek() < 48 || p.scanner.Peek() > 57 {
```
colincross: int('0') and int('9')

why not just p.accept(p.tok) and then check that the new p.tok is a scanner.Int?

parser/parser.go
Nan-Zhang: done

parser/ast.go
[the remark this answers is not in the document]
Nan-Zhang: thanks. so I just keep it as same as how bool, and string does for now

parser/parser.go
Nan-Zhang: done

parser/parser.go
Nan-Zhang: I think p.accept() allows scanner.comment and whitespace. since I don't hope any comments inserted in between "-" and int, so that's why I use scanner.Peek()

colincross: We accept comments almost anywhere between tokens, I don't see why this should be different.  You can enforce putting them after the in the printer package, which will fix them in bpfmt.

Nan-Zhang: ok. removed this check to just use p.accept().

mathjeff:
+1 looks good to me but probably should have Colin approve

PR #184, opened 2017-11-07
Disable bp finder

mathjeff:
waiting on some nonworking projects to be removed


mathjeff:
fixed all the problems; the reverts are now unneeded

PR #186, opened 2017-11-10
Don't allow user to define string, and bool properties within module struct

Nan-Zhang:
We only allow user to define string & bool properties with `blueprint:"mutated"` tag.


colincross:
I'm not sure this is something we want to enforce at the Blueprint level.  Using strings and bools is much simpler, and if you don't have complicated use cases that use append then *string and *bool are unnecessary.

PR #190, opened 2017-11-30
Have handleOneFile(child) wait for handleOneFile(parent).

mathjeff:
Bug: 65683273
Test: m -j nothing # which runs unit tests

Change-Id: I850e78ebcdbee68637f66d81fc127f19dd28508a

### DIFF
--- a/context.go
+++ b/context.go
@@ -585,9 +585,19 @@ func (c *Context) ListModulePaths(baseDir string) (paths []string, err error) {
 	return lines, nil
 }
 
-type stringAndScope struct {
-	string
-	*parser.Scope
+// a fileParseContext tells the status of parsing a particular file
+type fileParseContext struct {
+	// name of file
+	fileName string
+
+	// scope to use when resolving variables
+	Scope *parser.Scope
+
+	// pointer to the one in the parent directory
+	parent *fileParseContext
+
+	// is closed once FileHandler has completed for this file
+	doneVisiting chan struct{}
 }
 
 func (c *Context) ParseBlueprintsFiles(rootFile string) (deps []string, errs []error) {
@@ -629,29 +639,25 @@ func (c *Context) ParseFileList(rootDir string, filePaths []string) (deps []stri
 			return
 		}
 
-		atomic.AddInt32(&numGoroutines, 1)
-		go func() {
-			for _, def := range file.Defs {
-				var module *moduleInfo
-				var errs []error
-				switch def := def.(type) {
-				case *parser.Module:
-					module, errs = c.processModuleDef(def, file.Name)
-				case *parser.Assignment:
-					// Already handled via Scope object
-				default:
-					panic("unknown definition type")
-				}
-
-				if len(errs) > 0 {
-					atomic.AddUint32(&numErrs, uint32(len(errs)))
-					errsCh <- errs
-				} else if module != nil {
-					moduleCh <- module
-				}
+		for _, def := range file.Defs {
+			var module *moduleInfo
+			var errs []error
+			switch def := def.(type) {
+			case *parser.Module:
+				module, errs = c.processModuleDef(def, file.Name)
+			case *parser.Assignment:
+				// Already handled via Scope object
+			default:
+				panic("unknown definition type")
 			}
-			doneCh <- struct{}{}
-		}()
+
+			if len(errs) > 0 {
+				atomic.AddUint32(&numErrs, uint32(len(errs)))
+				errsCh <- errs
+			} else if module != nil {
+				moduleCh <- module
+			}
+		}
 	}
 
 	atomic.AddInt32(&numGoroutines, 1)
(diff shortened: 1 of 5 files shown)


context.go
```diff
@@ -623,41 +623,42 @@ func (c *Context) ParseFileList(rootDir string, filePaths []string) (deps []stri
 	var numErrs uint32
 	var numGoroutines int32
 
-	// handler must be reentrant
+	// handlers must be reentrant
+	fileCounter := func(file *parser.File) {
+		atomic.AddInt32(&numGoroutines, 1)
+	}
 	handleOneFile := func(file *parser.File) {
+
```
colincross: I generally prefer no blank line at the beginning of a function, unless the arguments to the function wraps to multiple lines.

context.go
```diff
@@ -702,11 +703,16 @@ type FileHandler func(*parser.File)
 // which the future output will depend is returned.  This list will include both
 // Blueprints file paths as well as directory paths for cases where wildcard
 // subdirs are found.
+//
+// syncVisitor will be called synchronously on each discovered file, and will only be called
```
colincross: This seems very abstract, especially since WalkBlueprintsFiles is only used once.  Maybe instead of adding syncVisitor just so you can pass a function to count goroutines, and since WalkBlueprintsFiles is now the one starting the goroutines, WalkBlueprintsFiles should count the goroutines itself and not exit until they all return?

context.go
```diff
@@ -778,7 +794,22 @@ loop:
 		case dep := <-depsCh:
 			deps = append(deps, dep)
 		case file := <-fileCh:
-			visitor(file)
+			syncVisitor(file)
+			go func() {
+				// wait for ancestor
+				<-visitCompletionChs[ancestorsMap[file.Name]]
```
colincross: Instead of looking up in two maps you could wrap the parser.File in a new type that holds a pointer to the ancestor (computed from descendantsMap) and a channel.  Then you wouldn't need ancestorsMap at all.

context.go
mathjeff: oops

context.go
mathjeff: sounds good

context.go
mathjeff: ok

context.go
```diff
@@ -586,8 +586,14 @@ func (c *Context) ListModulePaths(baseDir string) (paths []string, err error) {
 }
 
 type stringAndScope struct {
```
colincross: stringAndScope was always a bad name, but now it is even less accurate

context.go
```diff
@@ -702,6 +704,9 @@ type FileHandler func(*parser.File)
 // which the future output will depend is returned.  This list will include both
 // Blueprints file paths as well as directory paths for cases where wildcard
 // subdirs are found.
+//
+// visitor will be called asynchronously, and will only be called once visitor for each
+// ancestor directory has completed.
```
colincross: WalkBlueprintsFiles will not return until all calls to visitor have returned.

context.go
```diff
@@ -713,14 +718,13 @@ func (c *Context) WalkBlueprintsFiles(rootDir string, filePaths []string,
 	}
 	blueprintsSet := make(map[string]bool)
 
-	// Channels to receive data back from parseOneAsync goroutines
+	// Channels t	o receive data back from parseOneAsync goroutines
```
colincross: typo

context.go
```diff
@@ -731,18 +735,33 @@ func (c *Context) WalkBlueprintsFiles(rootDir string, filePaths []string,
...
 
-			doneCh <- blueprint
+			if blueprint.readyToVisit != nil {
+				// wait for visitor() of parent to complete
+				<-blueprint.readyToVisit
+			}
+
+			// process this file
+			visitor(file)
+			if blueprint.doneVisiting != nil {
```
colincross: IIUC doneVisiting is the same channel as readyToVisit in the descendants?  If so, that should be commented somewhere.

Coudl the blueprint object have a pointer to the parent blueprint object, and the readyToVisit check would become <-blueprint.parent.doneVisiting?

context_test.go
```diff
@@ -282,3 +284,46 @@ func createTestMutator(ctx TopDownMutatorContext) {
...
+			}
+		`),
+	}
+	ctx.MockFileSystem(mockFiles)
+
+	// prepare to monitor the visit order
+	visitOrder := []string{}
+	visitLock := sync.Mutex{}
+	visitor := func(file *parser.File) {
+		visitLock.Lock()
```
colincross: Might be worth adding a sleep or something in the parent to make sure the don't just happen to finish in the right order when running them asynchronously.

context.go
mathjeff: agreed

context.go
mathjeff: done

context.go
mathjeff: done

context.go
mathjeff: I've now implemented the latter

context_test.go
mathjeff: sounds good

mathjeff:
Assignee = Colin

context_test.go
colincross: I don't think a constant sleep will work, you need the goroutine that is started first to sleep longer than the one that is started second to ensure that the one that is started second can never run first.

colincross:
This needs to be rebased onto github's master, it's pulling in all the merge changes from AOSP

mathjeff:
oops; thanks

context_test.go
mathjeff: It turns out that the previous race condition inside WalkBlueprintsFiles applied to the order that visit was called, but that visit itself was called synchronously. As a result, adding some sleeps doesn't cause this test to fail (in practice) with the old code either.

Although I've now updated the sleep anyway.

PR #192, opened 2017-12-05
Fix propagating syntax errors to the user

mathjeff:


context.go
```diff
@@ -768,8 +768,10 @@ func (c *Context) WalkBlueprintsFiles(rootDir string, filePaths []string,
 				<-blueprint.parent.doneVisiting
 			}
 
-			// process this file
-			visitor(file)
+			if len(errs) < 1 {
```
colincross: len(errs) == 0

colincross:
I'm surprised this problem didn't cause any test failures.  Can you add a test that has an expected parse failure to make sure we exercise some of this code?

mathjeff:
Done

PR #194, opened 2017-12-09
Escape strings having spaces

mathjeff:
Bug: 70387174
Test: m -j out/soong/.intermediates/external/ltp/ltp_syscalls_h/gen/lapi/syscalls.h
Change-Id: I5231574bc01bb262ef142c8ef2fc4dc7f7b2bd74

### DIFF
--- a/proptools/escape.go
+++ b/proptools/escape.go
@@ -49,8 +49,7 @@ func ShellEscape(slice []string) []string {
 			r == '=',
 			r == '.',
 			r == ',',
-			r == '/',
-			r == ' ':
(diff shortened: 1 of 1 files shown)


mathjeff:
Failed

There are a bunch of cases where an Android.bp specifies cflags of for example "-include bionic-compat.h" which then results in command-line args of:

"-include bionic-compat.h"

which results in an error of

<built-in>:1:10: fatal error: ' bionic-compat.h' file not found
#include " bionic-compat.h"

PR #197, opened 2018-02-08
Create Document.md

123next:
This project uses conventional comments of languages like //,/* and #


123next:
This project uses conventional comments of  like //,/* and # which is a description of the comment type in this Project

PR #198, opened 2018-02-23
Fix glob cache conflict when excludes=nil and excludes=[]string{}

colincross:
Performing the same glob twice, once with excludes=nil and once
with excludes=[]string{} would hit the same entry in the glob
cache (since the glob filename would be the same), but fail
the verifyGlob check because DeepEqual considers []string(nil)
and []string{} to be different.  Use a manual array check
instead.

Test: glob_test.go
Change-Id: If0d4fe80163a871077b7276e1b4a3e888a4a4898

### DIFF
--- a/Blueprints
+++ b/Blueprints
@@ -23,6 +23,7 @@ bootstrap_go_package {
     ],
     testSrcs: [
         "context_test.go",
+        "glob_test.go",
         "ninja_strings_test.go",
         "ninja_writer_test.go",
         "splice_modules_test.go",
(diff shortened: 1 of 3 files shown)


Blueprints
```diff
@@ -23,6 +23,7 @@ bootstrap_go_package {
     ],
     testSrcs: [
         "context_test.go",
+	"glob_test.go",
```
danw: tab vs space

colincross: Fixed

PR #199, opened 2018-02-23
Fix format issues found by go vet

colincross:
Test: m checkbuild
Change-Id: I5135e0daecd20abce9a66631f5f55230168613de

### DIFF
--- a/bootstrap/bpglob/bpglob.go
+++ b/bootstrap/bpglob/bpglob.go
@@ -52,7 +52,7 @@ func (m *multiArg) Get() interface{} {
 }
 
 func usage() {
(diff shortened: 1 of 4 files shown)


glob.go
```diff
@@ -40,6 +40,14 @@ func verifyGlob(fileName, pattern string, excludes []string, g GlobPath) {
 }
 
 func (c *Context) glob(pattern string, excludes []string) ([]string, error) {
+	if excludes != nil && len(excludes) == 0 {
```
danw: This isn't a `go vet` change

colincross: Fixed

PR #200, opened 2018-02-23
Append / to directories in Glob results

danw:
This makes it easy for users of Glob to detect whether the match is a file or directory. Doing the check at this level means that the filelist file used as a dependency will be updated if a directory is replaced with a file of the same name, or vice versa.


danw:
I updated the godoc for pathtools.Glob, pathtools.GlobWithDeps, SingletonContext.GlobWithDeps, and ModuleContext.GlobWithDeps

PR #201, opened 2018-02-27
Support go1.10

danw:
Add stubs for the new testDeps interface functions. Also removes testing for 1.8.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1097,7 +1097,7 @@ func getLocalStringListFromScope(scope *parser.Scope, v string) ([]string, scann
 				Pos: assignment.EqualsPos,
 			}
 		default:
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -1097,7 +1097,7 @@ func getLocalStringListFromScope(scope *parser.Scope, v string) ([]string, scann
 				Pos: assignment.EqualsPos,
 			}
 		default:
-			panic(fmt.Errorf("unknown value type: %d", assignment.Value.Type))
+			panic(fmt.Errorf("unknown value type: %d", assignment.Value.Type()))
```
colincross: Is the go 1.10 compiler better about catching these?

danw: I think this is the `go vet` support in `go test`

PR #204, opened 2018-03-22
Add Patch and PatchList for making textual changes

colincross:
Patch and PatchList provide an API for making changes to substrings
of a Blueprint file by parsing the AST and using the token positions
to perform text replacements.

### DIFF
--- a/parser/modify_test.go
+++ b/parser/modify_test.go
@@ -1,0 +1,65 @@
+// Copyright 2018 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


parser/modify_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	err = patchlist.Add(12, 15, "x")
+	if _, ok := err.(PatchOverlapError); !ok {
+		t.Error("missing PatchOverlapError")
+	}
+
+	err = patchlist.Add(11, 15, "x")
+	if _, ok := err.(PatchOverlapError); !ok {
+		t.Error("missing PatchOverlapError")
+	}
```
danw: Can you test `patchlist.Add(15, 15, "x")`? That should return a PatchOverlapError, but if I'm reading the conditional correctly, will apply without error.

danw: Sorry, my complaint was that we weren't catching overlapping single character patches. Since we already did one patch on line 41, doing it again should fail with an error. But since we're doing a `<` in every section of the conditional, and both starts and ends are equal, we'll never error.

colincross: Done

PR #220, opened 2018-09-21
Improve glob behavior with symlinks

colincross:
Improve the testing around globs and symlinks.
Fix recursive glob walks through symlinks to directories.
Add an argument to pathtools.Glob to specify whether or not to follow symlinks.


pathtools/fs.go
```diff
@@ -220,8 +224,16 @@ func (m *mockFs) Exists(name string) (bool, bool, error) {
 }
 
 func (m *mockFs) IsDir(name string) (bool, error) {
+	origName := name
 	name = filepath.Clean(name)
 	name = m.followSymlinks(name)
+
+	dir := filepath.Dir(name)
+	dir = m.followSymlinks(dir)
```
danw: followSymlinks from 229 should have already done this, right?

pathtools/fs.go
```diff
@@ -220,8 +224,16 @@ func (m *mockFs) Exists(name string) (bool, bool, error) {
...
+	if _, ok := m.files[dir]; ok {
```
danw: Doesn't this check need to happen all the way up the tree?

pathtools/fs.go
colincross: I'll move this above m.followSymlinks(name).  It could be different if dir is a symlink and name is a symlink, for example:

foo/a -> a
bar -> foo

filepath.Dir(m.followSymlinks("bar/a")) will give "."
m.followSymlinks(filepath.Dir("bar/a")) will give "foo"

I'll try to add a test that covers this.

pathtools/fs.go
colincross: I'll try making IsDir recursive.

pathtools/fs.go
colincross: Fixed, and added new tests that would have incorrectly reported ENOENT but now report ENOTDIR.

PR #232, opened 2019-01-26
bpdoc improvements

colincross:
Extract module type documentation, and support primary builder customization of factory function to use for documentation for each module type.


colincross:
I'm going to refactor this.

colincross:
Refactored.

bootstrap/bpdoc/reader.go
```diff
@@ -0,0 +1,511 @@
...
+
+	if pkgPath != "" {
+		pkgPath += "/"
+	}
+	pkgPath += p[0]
+
+	return pkgPath, nil
+}
+
+func (r *Reader) ModuleType(name string, factory reflect.Value) (*ModuleType, error) {
```
gyias: I think it'd be more reader-friendly if we move this function to bpdoc.go. Having this only return a partially filled ModuleType instance and completing it in getModuleType feel like unneeded layering that only puzzles readers.

bootstrap/bpdoc/reader.go
```diff
@@ -0,0 +1,511 @@
...
+	}
+
+	return &ModuleType{
+		Name:    name,
+		PkgPath: pkgPath,
+		Text:    text,
+	}, nil
+}
+
+func (r *Reader) Package(pkgPath string) (*Package, error) {
```
gyias: In a similar vein, having this function here in reader.go when the definition of the Package struct itself and all the callers of this function are in bpdoc feels unnatural to me. The way I see it, reader.go is a sort of lower-level layer that handles more primitive actions, and making it handle higher-level structures blurs the border.

bootstrap/bpdoc/reader.go
```diff
@@ -0,0 +1,511 @@
...
+			}
+		}
+	}
+
+	return "", nil
+}
+
+// Return the PropertyStruct associated with a property struct type.  The type should be in the
+// format <package path>.<type name>
+func (r *Reader) PropertyStruct(pkgPath, name string, defaults reflect.Value) (*PropertyStruct, error) {
```
gyias: I have a similar feeling toward PropertyStruct-related functions, but have no clear idea how to better organize them. Would be willing to rearrange them myself once getting familiar with the code base, though.

PR #233, opened 2019-02-06
Support go 1.12 initialization functions

danw:
As listed in https://tip.golang.org/doc/go1.12#runtime

```
 Tracebacks, runtime.Caller, and runtime.Callers no longer include
 compiler-generated initialization functions. Doing a traceback during
 the initialization of a global variable will now show a function named
 PKG.init.ializers.
```

Test: run the soong tests with go 1.12 pre-release

### DIFF
--- a/package_ctx.go
+++ b/package_ctx.go
@@ -137,7 +137,7 @@ func checkCalledFromInit() {
 			panic("not called from an init func")
 		}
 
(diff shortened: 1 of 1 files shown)


danw:
Whoops, didn't test this as I thought I did

PR #236, opened 2019-02-15
Support examples as tests

colincross:
go test supports running examples as tests and verifying their
output, add support for running them to gotestmain.

Change-Id: If51abc64467d4701195cefc6db513f4d008794b5

### DIFF
--- a/gotestmain/gotestmain.go
+++ b/gotestmain/gotestmain.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"fmt"
 	"go/ast"
+	"go/doc"
 	"go/parser"
 	"go/token"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


gotestmain/gotestmain.go
```diff
@@ -121,6 +126,14 @@ var t = []testing.InternalTest{
 {{end}}
 }
 
+var e = []testing.InternalExample{
+{{range .Examples}}
+	{{if or .Output .EmptyOutput}}
+		{"{{.Name}}", pkg.Example{{.Name}}, {{.Output | printf "%q" }}, {{.Unordered}}},
+    {{end}}
```
gyias: The indentation seems a bit off here?

colincross: Fixed

PR #237, opened 2019-02-26
Allow Context to query Singletons

colincross:
Add Context.Singletons and Context.SingletonName to allow primary
builders to query Singletons returned by all registered
SingletonFactories.

Change-Id: Iee85643dd47f7472e6bb5ae8004374bd6ea0419e

### DIFF
--- a/singleton_ctx.go
+++ b/singleton_ctx.go
@@ -27,6 +27,8 @@ type Singleton interface {
 type SingletonContext interface {
 	Config() interface{}
 
+	Name() string
+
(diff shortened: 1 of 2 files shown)


singleton_ctx.go
```diff
@@ -27,6 +27,8 @@ type Singleton interface {
 type SingletonContext interface {
 	Config() interface{}
 
+	SingletonName() string
```
gyias: How about just Name()?

colincross: Done

PR #239, opened 2019-02-28
Support go 1.12

colincross:


package_ctx.go
```diff
@@ -137,12 +138,17 @@ func checkCalledFromInit() {
 			panic("not called from an init func")
 		}
 
-		if funcName == "init" || strings.HasPrefix(funcName, "init·") {
+		if funcName == "init" || strings.HasPrefix(funcName, "init·") ||
```
asmundak: Why is it checking whether `funcName` starts with 'init' followed by bullet character?

colincross: Go sometimes uses "middle dot" as a character in function names.  Looks like it hasn't done that at least for init functions since https://go-review.googlesource.com/c/go/+/3964/

PR #240, opened 2019-03-01
Add more OtherModule* methods

colincross:
Add OtherModuleDir, OtherModuleSubDir, and OtherModuleType to
ModuleContext and TopDownMutatorContext, and add ModuleType to
BaseModuleContext.

Change-Id: If5c873a2620ef10333c0bdba5ab89d4256e5fdf2

### DIFF
--- a/module_ctx.go
+++ b/module_ctx.go
@@ -123,6 +123,7 @@ type DynamicDependerModule interface {
 type BaseModuleContext interface {
 	ModuleName() string
 	ModuleDir() string
+	ModuleType() string
 	Config() interface{}
 
 	ContainsProperty(name string) bool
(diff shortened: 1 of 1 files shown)


module_ctx.go
```diff
@@ -154,6 +155,9 @@ type ModuleContext interface {
 	BaseModuleContext
 
 	OtherModuleName(m Module) string
+	OtherModuleDir(m Module) string
```
asmundak: 1. Would it be better to have a single function returning struct than five OtherModuleXxxx functions with same arguments?
2. Please document what these functions do.

colincross: 1. I don't think so, most of the callers only want a single value.
2.  I'll do that in a follow-up, the whole interfaces need to be documented, not just these new functions.

PR #252, opened 2019-06-19
Put the primary builder into the console pool

colincross:
The primary builder should always be running on its own with no
other processes running in parallel, so put it in the console
pool so that it can print to stdout without buffering.

Bug: 80165685
Test: build with print in Soong
Change-Id: If34ecdb5fa18de7e47c4cd6965d551c504850176

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -130,6 +130,7 @@ var (
 			Deps:        blueprint.DepsGCC,
 			Depfile:     "$out.d",
 			Restat:      true,
+			Pool:        blueprint.Console,
 		},
 		"builder", "extra", "generator", "globFile")
 


danw:
I don't intend to support console for any ninja replacement, and am worried about synchronization issues between the protobuf frontend and stdout. Why is there any time between printing an error and actually exiting?

(I have some cases that run the primary builder in parallel with other actions -- a ninja-level `checkbuild` type)

colincross:
With https://android-review.googlesource.com/c/platform/build/soong/+/986046, at startup soong_build prints:
starting dlv attach 166432 --headless -l :1234 --api-version=2 --accept-multiclient --log
API server listening at: [::]:1234
time="2019-06-19T15:21:37-07:00" level=info msg="attaching to pid 166432" layer=debugger

I'd like that to be visible, otherwise it is unclear when to connect the debugger, and if you forget that you set SOONG_DELVE it will pause there forever.

PR #259, opened 2019-08-09
Add SetDefaultDependencyVariation

colincross:
SetDefaultDependencyVariation sets the variation name that will be used
when a dangling dependency is found while a module is being split. A
dangling dependency can occur if a module is split to a variant that one
of its dependencies is not split into. When the default variation is not
set, such dangling dependency is a hard error. But with the new
function, the default variation can be set and subsequent calls to
CreateVariations and its variations on the same context uses the default
variation when necessary.

(If even the default variation does not exist for the dependent module,
it is an hard error)

Note that this is different from calling SetDependencyVariation("foo")
followed by CreateVariations("foo", "bar"). In that case, regardless of
whether a dependency of the current module has the variant 'bar' or not,
only the 'foo' variant is chosen.

With SetDefaultDependencyVariation("foo") followed by
CreateVariations("foo", "bar"), 'foo' variant is used only when the
'bar' variant of the current module depends on a module that does not
have 'bar' variant.

Bug: 138103882
Test: m
Change-Id: I4520ca87487994de024fdbacda3bef6636225f0d

### DIFF
--- a/context.go
+++ b/context.go
@@ -1153,7 +1153,7 @@ func (c *Context) cloneLogicModule(origModule *moduleInfo) (Module, []interface{
 }
 
 func (c *Context) createVariations(origModule *moduleInfo, mutatorName string,
-	variationNames []string) ([]*moduleInfo, []error) {
+	defaultVariationName *string, variationNames []string) ([]*moduleInfo, []error) {
 
 	if len(variationNames) == 0 {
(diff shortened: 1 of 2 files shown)


colincross:
Reviewed at https://android-review.googlesource.com/c/platform/build/blueprint/+/1093129

PR #260, opened 2019-09-05
Fix checkCalledFromInit

jooyunghan:
This function checks if it is called from init() by looking into
callers. By the way, it may fail when init() is inlined.

To fix this, CallersFrames() is used to translate PCs into symbolic
information accounting for inlined functions.

Test: go test ./...
Change-Id: I18b3c3ffdaf71f775e3522c87607aec5df7b09c5

### DIFF
--- a/package_ctx.go
+++ b/package_ctx.go
@@ -158,8 +158,9 @@ func callerName(skip int) (pkgPath, funcName string, ok bool) {
 	if n != 1 {
 		return "", "", false
 	}
-
-	f := runtime.FuncForPC(pc[0]).Name()
+	frames := runtime.CallersFrames(pc[:])
+	frame, _ := frames.Next()
+	f := frame.Function
 	s := pkgPathRe.FindStringSubmatch(f)
 	if len(s) < 3 {
 		panic(fmt.Errorf("failed to extract package path and function name from %q", f))


jooyunghan:
@googlebot I signed it!

danw:
Hmm, I wondered if this was a go1.12 vs go1.13 difference (since I wasn't reproducing it), but trying on go1.13rc1 doesn't reproduce either.

jooyunghan:
I ran into this with "go test" in "<aosp>/build/soong/apex" and "go version" shows "go version go1.13rc1 linux/amd64" on my local machine.

```
$ go test
panic: not called from an init func

goroutine 1 [running]:
github.com/google/blueprint.checkCalledFromInit()
	~/work/aosp-master/build/blueprint/package_ctx.go:138 +0x164
github.com/google/blueprint.(*packageContext).StaticVariable(0xc0001ec0f0, 0x9f8696, 0xd, 0xc000024940, 0x3d, 0xc000024940, 0x3d)
	~/work/aosp-master/build/blueprint/package_ctx.go:267 +0x26
android/soong/cc/config.init.0()
	~/work/aosp-master/build/soong/cc/config/arm64_device.go:95 +0x1b3
exit status 2
FAIL	android/soong/apex	0.075s
```

danw:
Ah, in build/soong/apex. Yep, seems like a go 1.12 vs 1.13 issue (I can replicate it on 1.13, but not 1.12)

PR #267, opened 2019-10-17
Move outputs of bootstrap modules in main stage to separate directory

colincross:
Don't write to $buildDir/.bootstrap in the ninja file generated for
the main stage.

Test: examine main stage build.ninja file
Change-Id: I9c6183c9a434ff11d44cdd4f9844eae192ebfe03

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -27,6 +27,7 @@ import (
 	"github.com/google/blueprint/pathtools"
 )
 
+const mainSubDir = ".primary"
 const bootstrapSubDir = ".bootstrap"
 const miniBootstrapSubDir = ".minibootstrap"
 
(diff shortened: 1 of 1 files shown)


buchgr:
Thank you @colincross.

Do I understand correctly that `StageMain` is the bootstrapping stage where the blueprint project (ie soong) gets built and `StagePrimary` is the actual build stage? This code changes it so that during `StageMain` everything is written in $OUT_DIR/.primary and in `StagePrimary` files are written in $OUT_DIR/.bootstrap. Is this accurate?

colincross:
The existing naming here was a little confusing.  `StagePrimary` is the stage that is _building_ the primary builder, and `StageMain` is the stage that is _running_ the primary builder.  I left the files used to build the primary builder in `.bootstrap`, and then moved the files built by the primary builder to `.primary`.

danw:
Even that's a bit ambiguous...

`StagePrimary` is where we build and run the primary builder (soong_build for Android) and generate the action graph for the "Main" stage.

`StageMain` is where we run the vast majority of the build, and that largely consists of the custom rules defined by the primary builder, in addition to the handful of binary go tools (and their associated packages) that this package defines.

buchgr:
thanks @colincross and @danw !

PR #268, opened 2019-10-18
Two examples of Blueprint usage.

skvadrik:
- An example of build.ninja generator that shows Blueprint "stages"
  and the relations between Blueprint concept of modules and Ninja
  build input/output targets.

- An example of a custom Blueprint-based bootstrapping build system
  ("simplebp" written by Tim Kilbourn). Note that it adds Blueprint
  source code as a submodule.


skvadrik:
Eh, I see test failures (probably due to not masking Blueprints files in examples subdirectory). Will clean them up and resubmit (I thought `go test` does all the checking).

skvadrik:
I fixed `go test ./...` failures by renaming `examples` into `.examples` so that blueprint_impl.bash:14 does not find the Blueprints files from the examples.

skvadrik:
Any update on this (before it bitrots)?

PR #270, opened 2019-11-18
Slice properties can be replaced

jiyongp:
override_* in Soong requires a module to override certain properties of
other module. In that case, values of a slice property (e.g. []string)
should be replaced by the same property value in the overriding module.
However, since proptools only supports Append and Prepend orders where
the original values are kept for slice properties, the behavior
couldn't be implemented. To support the use case, Replace order is
introduced, in which case slice property values are completely replaced.
For other types of properties, the Replace order behaves exactly the
same as the Append order.

Bug: 144338929
Test: m

Change-Id: Iae9feda035177fe6a22e6e8319c0fdaa9e08e85e

### DIFF
--- a/proptools/extend.go
+++ b/proptools/extend.go
@@ -150,6 +150,7 @@ type Order int
 const (
 	Append Order = iota
 	Prepend
+	Replace
 )
 
 type ExtendPropertyFilterFunc func(property string,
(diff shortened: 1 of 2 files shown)


colincross:
Needs gofmt

jiyongp:
@colincross, I've just ran gofmt
@gyias That is correctly handled. I've added a test case for that.

jiyongp:
Thanks Jaewoong. Do you plan to upload the merge in AOSP? (or should I?)

gyias:
Feel free to go ahead and send a merge CL to me if you know how. Let me know if you need help.

jiyongp:
Done.

PR #271, opened 2019-12-10
prettyPrint missing mutators

jooyunghan:
prettyPrintVariant() prints only for the known mutators. But when there
are unknown mutators they are missing from the error message. This leads
to somewhat wierd message like following:

dependency "foo" of "bar" missing variant:
  A:a, B:b
available variants:
  A:a, B:b

This can happend when we call AddVariationDependencies() like following:
  ctx.AddVariationDependencies([]Variation{
  	{"A", "a"},
  	{"B", "b"},
  	{"C", "c"},
  }, ...)
when there are only "A" and "B" registered.

By printing missing variants, it will print like following:

dependency "foo" of "bar" missing variant:
  A:a, B:b (C:c)
available variants:
  A:a, B:b

### DIFF
--- a/context.go
+++ b/context.go
@@ -1272,8 +1272,20 @@ func (c *Context) prettyPrintVariant(variant variationMap) string {
 			names = append(names, m+":"+v)
 		}
 	}
-
-	return strings.Join(names, ", ")
+	var missing string
+	if len(names) < len(variant) {
+		variantClone := variant.clone()
+		for _, m := range c.variantMutatorNames {
+			delete(variantClone, m)
+		}
+		unknownVariants := make([]string, 0, len(variant)-len(names))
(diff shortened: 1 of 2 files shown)


colincross:
How are you ending up with missing mutators?  Maybe it should be an error to pass in a variation with a mutator that is not in the list of known mutators?

jooyunghan:
While writing tests for aidl_interface, I copied required dependencies of modules types and mutators. When I put several required ndk_library modules I got errors like that.
```
dependency "foo" of "bar" missing variant:
A:a, B:b
available variants:
A:a, B:b
```

Emitting an error of something like "unknown mutator blah blah.." would be better than this PR.
like..

In any way, the current error message doesn't help.

PR #277, opened 2020-01-16
Speed up incremental builds after globs

colincross:
Two issues with globs were causing unnecessary primary builder reruns on an incremental build after a clean build or after changing a glob result.

### DIFF
--- a/bootstrap/glob.go
+++ b/bootstrap/glob.go
@@ -131,8 +131,14 @@ func (s *globSingleton) GenerateBuildActions(ctx blueprint.SingletonContext) {
 			depFile := fileListFile + ".d"
 
 			fileList := strings.Join(g.Files, "\n") + "\n"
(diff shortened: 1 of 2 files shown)


bootstrap/glob.go
```diff
@@ -131,8 +131,8 @@ func (s *globSingleton) GenerateBuildActions(ctx blueprint.SingletonContext) {
 			depFile := fileListFile + ".d"
 
 			fileList := strings.Join(g.Files, "\n") + "\n"
-			pathtools.WriteFileIfChanged(fileListFile, []byte(fileList), 0666)
-			deptools.WriteDepFile(depFile, fileListFile, g.Deps)
+			pathtools.WriteFileIfChanged(absolutePath(fileListFile), []byte(fileList), 0666)
```
danw: Why wasn't this triggering an error that we'd notice?

colincross: Because it doesn't check the returned error.  I'll fix that too.

PR #291, opened 2020-03-05
bpmodify: fix os.Exit() shouldn't shadow panic()

silverneko:
The `defer func() { os.Exit() }()` in main() method shadows panic().
Make the exit handler recover() from panic(), log the panic(), and then
gracefully exit.

Test: m bpmodify
Change-Id: Icc89f8fce0b6096489baa0ba0f08c21d1ef623bc

### DIFF
--- a/bpmodify/bpmodify.go
+++ b/bpmodify/bpmodify.go
@@ -223,7 +223,12 @@ func walkDir(path string) {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


silverneko:
Hot fix!
I just found out that my previous change has unwanted side-effects of shadowing a panic() call.

PR #292, opened 2020-03-24
Allow missing variants when allowMissingDependencies=true

paulduffin:


paulduffin:
@colincross Would you be able to pull this change?

PR #294, opened 2020-04-02
bpmodify: handle nested properties

silverneko:
aosp/1250007

Add an option "-property", which is an alias to the option "-parameter".

For example if Android.bp contains:

  cc_foo {
    name: "foo",
  }

Then `bpmodify -m foo -a bar -property baz.buz Android.bp` outputs:

  cc_foo {
    name: "foo",
    baz: {
      buz: ["bar"],
    },
  }

Bug: 149715904
Test: go test -v
Change-Id: I9660cff1b5239ccf5aa9ef1a41835b8ac6cd4b9f

### DIFF
--- a/bpmodify/bpmodify.go
+++ b/bpmodify/bpmodify.go
@@ -22,18 +22,20 @@ import (
 
 var (
 	// main operation modes
-	list            = flag.Bool("l", false, "list files that would be modified by bpmodify")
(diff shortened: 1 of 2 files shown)


danw (approved):
At some point we should probably rewrite the rest of this tool to work like bpfix and do patch updates, which is somewhat better at preserving existing formatting. But that's out of scope to this change.

PR #295, opened 2020-04-06
Add String() method to pretty print BaseDependencyTag

paulduffin:


paulduffin:
@colincross is this what you meant?

paulduffin:
This is useless. It will only ever print the contents of the BaseDependencyTag structure.

PR #297, opened 2020-05-04
Run LoadHooks before registering module

paulduffin:
Previously a LoadHook could not modify the name of a module because the
module was registered before the LoadHooks were run. That made it very
complicated (requiring mutators and auto generated names) to create a
module type whose name was determined by say the directory in which it
is defined.

This change moves the LoadHook execution slightly earlier so it runs
before registration of the module.

That caused one slight side problem which was that the
LoadHookContext.ModuleName() method would fail when called in a
LoadHook. That was because ModuleName() calls *moduleInfo.Name() which
gets the name from moduleInfo.group.name but that resulted in a panic
as group was nil because it is only set when the module is registered.

Overriding the ModuleName() method to get the name from the module
itself (using moduleInfo.logicModule.Name()) fixed that. The reason for
getting the name from the module.group rather than the module is that
the name may have been changed however that is not an issue in this
case as there has been no opportunity for the module to be renamed.

### DIFF
--- a/context.go
+++ b/context.go
@@ -218,7 +218,16 @@ type depInfo struct {
 }
 
 func (module *moduleInfo) Name() string {
-	return module.group.name
+	// If this is called from a LoadHook (which is run before the module has been registered)
+	// then group will not be set and so the name is retrieved from logicModule.Name().
+	// Usually, using that method is not safe as it does not track renames (group.name does).
+	// However, when called from LoadHook it is safe as there is no way to rename a module
+	// until after the LoadHook has run and the module has been registered.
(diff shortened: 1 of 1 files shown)


paulduffin:
I found an issue in some more methods called on LoadHookContext, e.g. PropertyErrorf(), ModuleErrorf() all of which end up calling moduleInfo.String() which calls moduleInfo.Name() which fails because group is nil. So, I removed the override of ModuleName() and instead made moduleInfo.Name() check if group is nil and return logicModule.Name() instead. That fixes those and any other similar issues.

PR #299, opened 2020-05-13
Support checking syntax of generated Blueprint files

paulduffin:
Adds a CheckBlueprintSyntax(...) method to check the syntax of a
Blueprint file.

Changes processModuleDef and newModule from being method on *Context to
being standalone functions. That ensures that CheckBlueprintSyntax(...)
does not need to take a context and so there is no chance that it can
change its state.


paulduffin:
Ping.

PR #300, opened 2020-05-15
Add ctx.OtherModule(Reverse)DependencyVariantExists.

marstj:
This allows guarding calls to AddVariationDependencies and
AddReverseDependency to register dependencies on optional modules.

### DIFF
--- a/module_ctx.go
+++ b/module_ctx.go
@@ -275,6 +275,19 @@ type BaseModuleContext interface {
 	// OtherModuleExists returns true if a module with the specified name exists, as determined by the NameInterface
 	// passed to Context.SetNameInterface, or SimpleNameInterface if it was not called.
 	OtherModuleExists(name string) bool
+
+	// OtherModuleDependencyVariantExists returns true if a module with the
+	// specified name and variant exists. The variant must match the given
+	// variations. It must also match all the non-local variations of the current
+	// module. In other words, it checks for the module AddVariationDependencies
+	// would add a dependency on with the same arguments.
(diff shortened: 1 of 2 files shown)


module_ctx.go
```diff
@@ -275,6 +275,17 @@ type BaseModuleContext interface {
...
+	// OtherModuleVariantExists returns true if a module with the specified name
+	// and variant exists. The variant must match the given variations. It must
+	// also match all variations of the current module if reverse is true, or all
+	// the non-local variations of the current module if reverse is false.
+	//
+	// In other words, if reverse is false it checks for the module
+	// AddVariationDependencies would add a dependency on with the same arguments.
+	// If reverse is true and variations is nil, it checks for the module
+	// AddReverseDependency would add a dependency on.
+	OtherModuleVariantExists(variations []Variation, name string, reverse bool) bool
```
colincross, line 288: I'd prefer not to take the "reverse" argument.  Can you instead add two separate methods that call a helper with reverse true or false under the hood?  Something like OtherModuleDependencyVariantExists and OtherModuleReverseDependencyVariantExists.

marstj: Ok. Should OtherModuleReverseDependencyVariantExists take a variations argument then, seeing as the corresponding AddReverseDependency doesn't do that?

colincross: Leaving out the variation sounds good for OtherModuleReverseDependencyVariantExists

marstj: Done, PTAL. I've tested the change with my use case in https://r.android.com/q/topic:prebuilt-variants+author:mast.

PR #304, opened 2020-06-15
Export ModuleListFile in bootstrap (with fix)

c-parsons:


bootstrap/command.go
```diff
@@ -41,10 +41,10 @@ var (
 	traceFile      string
 	runGoTests     bool
 	noGC           bool
-	moduleListFile string
 	emptyNinjaFile bool
 
 	BuildDir      string
+	ModuleListFile string
```
danw, line 47: This is failing in travis because gofmt needs to be run. (Travis still isn't reporting on pull requests... I'll fix that by moving us to github actions later today)

c-parsons: Done. Thanks!

PR #310, opened 2020-07-09
Add extensions to context.go for introspection features

jin:
See https://android-review.googlesource.com/c/platform/build/soong/+/1359245 for context

### DIFF
--- a/context.go
+++ b/context.go
@@ -217,14 +217,6 @@ type depInfo struct {
 	tag    DependencyTag
 }
 
-func (c *Context) LogicModules() []Module {
-	modules := []Module{}
-	for k, _ := range c.moduleInfo {
-		modules = append(modules, k)
-	}
-	return modules
-}
-
 func (module *moduleInfo) Name() string {
 	// If this is called from a LoadHook (which is run before the module has been registered)
 	// then group will not be set and so the name is retrieved from logicModule.Name().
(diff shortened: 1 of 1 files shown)


paulduffin:
The TargetNameWithVariant() and QualifiedTargetLabel() methods do not appear to need to be in Blueprint as they only use capabilities that are available through the existing interfaces. They are also quite specific to the use being made of them inside Soong (the Android build system) and so should be part of that not Blueprint. You will also find that there is already some similar code used in the visibility functionality in Soong which was based, at least in part, on the Bazel support for visibility.

jin:
@paulduffin Thank you! Good idea.

context.go
```diff
@@ -3071,6 +3071,22 @@ func (c *Context) BlueprintFile(logicModule Module) string {
 	return module.relBlueprintsFile
 }
 
+func (c *Context) LogicModules() []Module {
```
colincross, line 3074: This is already available through ctx.VisitAllModules.

context.go
```diff
@@ -3071,6 +3071,22 @@ func (c *Context) BlueprintFile(logicModule Module) string {
...
+	modules := []Module{}
+	for k, _ := range c.moduleInfo {
+		modules = append(modules, k)
+	}
+	return modules
+}
+
+func (c *Context) DirectDeps(logicModule Module) []Module {
```
colincross, line 3082: This is already available through ctx.VisitDrirectDeps.

context.go
jin: Ahh, of course! Thank you, this PR is not needed then :)

PR #312, opened 2020-07-16
Update bpdocs to filter nested properties by tags

lizkammer:
Prior to this change nested properties tagged with `blueprint:"mutated"`
are erroneously included in documentation, this corrects that behavior
and adds testing to verify.

### DIFF
--- a/bootstrap/bpdoc/bpdoc_test.go
+++ b/bootstrap/bpdoc/bpdoc_test.go
@@ -44,3 +44,34 @@ func TestNestedPropertyStructs(t *testing.T) {
 		}
 	}
 }
+
+func TestAllModules(t *testing.T) {
+	packages, err := AllPackages(pkgFiles, moduleTypeNameFactories, moduleTypeNamePropertyStructs)
+	if err != nil {
+		t.Fatalf("expected no errors, got %s", err)
+	}
+
+	if numPackages := len(packages); numPackages != 1 {
+		t.Fatalf("Expected 1 package, got %q", numPackages)
+	}
+
+	p := packages[0]
+
+	for _, m := range p.ModuleTypes {
+		for _, p := range m.PropertyStructs {
+			noMutatedProperties(t, p.Properties)
+		}
+	}
+
+}
+
+func noMutatedProperties(t *testing.T, properties []Property) {
+	t.Helper()
+
+	for _, p := range properties {
+		if hasTag(p.Tag, "blueprint", "mutated") {
+			t.Errorf("Property %s has `blueprint:\"mutated\" tag but should have been excluded.", p)
+		}
+		noMutatedProperties(t, p.Properties)
+	}
+}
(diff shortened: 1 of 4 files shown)


bootstrap/bpdoc/bpdoc_test.go
```diff
@@ -44,3 +44,34 @@ func TestNestedPropertyStructs(t *testing.T) {
...
 }
+
+func TestAllPackages(t *testing.T) {
+	packages, err := AllPackages(pkgFiles, moduleTypeNameFactories, moduleTypeNamePropertyStructs)
+	if err != nil {
+		t.Fatalf("expected no errors, got %s", err)
+	}
+
+	if numPackages := len(packages); numPackages != 1 {
+		t.Fatalf("Expected 1 package, got %q", numPackages)
```
gyias, line 55: Wouldn't it be more helpful if this prints out the actual contents of packages when it fails?

bootstrap/bpdoc/bpdoc_test.go
```diff
@@ -44,3 +44,34 @@ func TestNestedPropertyStructs(t *testing.T) {
 		}
 	}
 }
+
+func TestAllPackages(t *testing.T) {
+	packages, err := AllPackages(pkgFiles, moduleTypeNameFactories, moduleTypeNamePropertyStructs)
+	if err != nil {
+		t.Fatalf("expected no errors, got %s", err)
```
patricearruda, line 51: Just noticed this: I see that you are using t.Fatalf. t.Fatalf is used mostly for setting up test cases. This looks like a validation step. You should use t.Errorf and let the test case continue since it might discover additional issues. You should be more descriptive about your error:

t.Errorf("got %v, expecting nil error on retrieving all <something> about packages", err)

bootstrap/bpdoc/bpdoc_test.go
```diff
@@ -44,3 +44,34 @@ func TestNestedPropertyStructs(t *testing.T) {
...
+
+	if numPackages := len(packages); numPackages != 1 {
+		t.Fatalf("Expected 1 package, got %q", numPackages)
+	}
+
+	p := packages[0]
+
+	for _, m := range p.ModuleTypes {
+		for _, p := range m.PropertyStructs {
+			noMutatedProperties(t, p.Properties)
```
patricearruda, line 62: This looks like an assert function. In go readability, assert functions is not recommended on test practices (see go/gocomments#assert). I would recommend  the function notMutatedProperties to return back an error and let the caller decide how to handle the error.

bootstrap/bpdoc/bpdoc_test.go
lizkammer: Done.

bootstrap/bpdoc/bpdoc_test.go
lizkammer: Done.

bootstrap/bpdoc/bpdoc_test.go
lizkammer: Done.

PR #316, opened 2020-08-21
Return dependency modules from dependency-adding methods.

skvadrik:
This is the first part of the change: it adds a shadow method (suffixed
with 'X') for each existing method that adds dependencies. Existing
methods call the corresponding shadow methods and discard return value.

The change does not add runtime overhead on findng dependency modules,
as this has already been done previously.

Test: go test

---
If this approach is ok, further steps will be:

- add similar shadow methods in Soong
- port all Soong code to use the new methods
- remove original (now unused) methods from Soong
- remove original (now unused) methods from Blueprint
- rename methods to get rid of X in Soong
- rename methods to get rid of X in Blueprint

I thought it might be easier to do in many small steps considering how invasive the change is.

### DIFF
--- a/module_ctx.go
+++ b/module_ctx.go
@@ -775,10 +775,15 @@ type TopDownMutatorContext interface {
 type BottomUpMutatorContext interface {
 	BaseMutatorContext
 
-	// AddDependency adds a dependency to the given module.
-	// Does not affect the ordering of the current mutator pass, but will be ordered
-	// correctly for all future mutator passes.
-	AddDependency(module Module, tag DependencyTag, name ...string)
+	// AddDependency adds a dependency to the given module.  It returns a slice of modules for each
+	// dependency (some entries may be nil).  Does not affect the ordering of the current mutator
+	// pass, but will be ordered correctly for all future mutator passes.
+	//
+	// If the mutator is parallel (see MutatorHandle.Parallel), this method will pause until the
+	// new dependencies have had the current mutator called on them.  If the mutator is not
+	// parallel this method does not affect the ordering of the current mutator pass, but will
+	// be ordered correctly for all future mutator passes.
+	AddDependency(module Module, tag DependencyTag, name ...string) []Module
 
 	// AddReverseDependency adds a dependency from the destination to the given module.
 	// Does not affect the ordering of the current mutator pass, but will be ordered
@@ -818,19 +823,30 @@ type BottomUpMutatorContext interface {
 	SetDefaultDependencyVariation(*string)
 
 	// AddVariationDependencies adds deps as dependencies of the current module, but uses the variations
-	// argument to select which variant of the dependency to use.  A variant of the dependency must
-	// exist that matches the all of the non-local variations of the current module, plus the variations
-	// argument.
-	AddVariationDependencies([]Variation, DependencyTag, ...string)
+	// argument to select which variant of the dependency to use.  It returns a slice of modules for
+	// each dependency (some entries may be nil).  A variant of the dependency must exist that matches
+	// the all of the non-local variations of the current module, plus the variations argument.
+	//
+	// If the mutator is parallel (see MutatorHandle.Parallel), this method will pause until the
+	// new dependencies have had the current mutator called on them.  If the mutator is not
+	// parallel this method does not affect the ordering of the current mutator pass, but will
+	// be ordered correctly for all future mutator passes.
+	AddVariationDependencies([]Variation, DependencyTag, ...string) []Module
 
 	// AddFarVariationDependencies adds deps as dependencies of the current module, but uses the
-	// variations argument to select which variant of the dependency to use.  A variant of the
-	// dependency must exist that matches the variations argument, but may also have other variations.
+	// variations argument to select which variant of the dependency to use.  It returns a slice of
+	// modules for each dependency (some entries may be nil).  A variant of the dependency must
+	// exist that matches the variations argument, but may also have other variations.
 	// For any unspecified variation the first variant will be used.
 	//
 	// Unlike AddVariationDependencies, the variations of the current module are ignored - the
 	// dependency only needs to match the supplied variations.
-	AddFarVariationDependencies([]Variation, DependencyTag, ...string)
+	//
+	// If the mutator is parallel (see MutatorHandle.Parallel), this method will pause until the
+	// new dependencies have had the current mutator called on them.  If the mutator is not
+	// parallel this method does not affect the ordering of the current mutator pass, but will
+	// be ordered correctly for all future mutator passes.
+	AddFarVariationDependencies([]Variation, DependencyTag, ...string) []Module
 
 	// AddInterVariantDependency adds a dependency between two variants of the same module.  Variants are always
 	// ordered in the same orderas they were listed in CreateVariations, and AddInterVariantDependency does not change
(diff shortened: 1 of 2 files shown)


skvadrik:
> Are the X methods necessary?

Got rid of them after realizing that Go allows to ignore return values. This simplifies the patch considerably and makes the extra-steps unnecessary.

> This returns a reference to a module that is in an undefined state w.r.t. the mutator, and may be being modified concurrently. If we could delay returning from AddVariationDependencies until the mutator pass on the new dependency has finished (watching out for cycles, which would cause a deadlock) then this would be much more consistent.

I considered possible implementation of this idea for a while, and I can't find an efficient approach.

1. Waiting until the dependency module has been processed by the current mutator is not difficult. It can probably be implemented with a condition variable for each module, where the mutator wakes all the blocked threads when it ends visiting that module.

2. Ensuring no deadlocks in a static graph is worst-case linear in the size of the graph. It boils down to checking if adding an edge to a DAG will introduce a loop. Checking for a loop can be done with a DFS or BFS, which is O(n) where n is the number of nodes (modules), and if one has a topological ordering of nodes it is possible to cheat by checking in O(1) if the newly added edge goes from earlier index to a later index --- in that case adding the edge won't break the topological order and there's no need to do the O(n) check. All in all, not cheap, but feasible. It makes sense to add a specialized method for that rather than adding the overhead to default methods which will be unused most of the time.

3. Ensuing no deadlocks in a dynamic graph is possible, but it defeats the purpose of the parallel mutator. Since the graph is changing (new edges are added) in parallel with the deadlock analysis, it is not possible to use the precomputed topological ordering. What is even worse, it is hardly possible to do any deadlock analysis without locking the full graph (what if a new edge is added in parallel that will create a loop?). Locking the full graph will likely make the mutator slower than sequential.

skvadrik:
Maybe we can return a view of the immutable part of the module?

colincross:
Unfortunately Go doesn't really have a concept of immutability, so there isn't an immutable part of the module to return a view of.

I have some ideas on how to detect cycles without overhead, I can work on it once I finish some other Blueprint fixes.  The controller inside parallelVisit has visibility into how many are actually running in order to enforce the parallelism limit.  When new dependencies are added, the goroutine that is waiting needs to communicate back with the controller via a channel to say it is paused on a module in order to be resumed when the other module is done.  The controller can detect the case of zero runnable modules while at at least one module is paused or unstarted, and then run the expensive cycle detector to report why.

skvadrik:
Cool! This is a much better approach. It reminds me of Grace Hopper's _"It's easier to ask forgiveness than it is to get permission"_.

module_ctx.go
```diff
@@ -998,57 +998,73 @@ func (mctx *mutatorContext) Module() Module {
...
+	depInfos := make([]Module, 0, len(deps))
 	for _, dep := range deps {
 		modInfo := mctx.context.moduleInfo[module]
-		errs := mctx.context.addDependency(modInfo, tag, dep)
+		depInfo, errs := mctx.context.addDependency(modInfo, tag, dep)
 		if len(errs) > 0 {
 			mctx.errs = append(mctx.errs, errs...)
 		}
+		if depInfo != nil {
+			depInfos = append(depInfos, depInfo.logicModule)
```
colincross, line 1010: This is kind of odd behavior, the length of the return value is different than the length of the deps in the case of missing dependencies.  This seems likely to cause common panics for missing dependencies, for example:

```
dep := mctx.AddDependency("foo")[0] // panic here due to an index out of range error
transitive = append(transitive, dep.transitive...)
```

The alternative would be to return nils in the slice for missing dependencies, but that just delays the problem:
```
dep := mctx.AddDependency("foo")[0]
transitive = append(transitive, dep.transitive...) // panic here due to nil pointer dererference
```

Either way is going to need a bunch of error checking by anyone who uses the return value.  I think the second is a little easier to reason about though.

module_ctx.go
```diff
@@ -778,14 +778,14 @@ type BottomUpMutatorContext interface {
 	// AddDependency adds a dependency to the given module.
 	// Does not affect the ordering of the current mutator pass, but will be ordered
 	// correctly for all future mutator passes.
```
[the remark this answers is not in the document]
skvadrik: Done.

module_ctx.go
skvadrik: It is slightly more convenient to iterate over a shorter or empty slice than over a slice with nils, but I agree that returning nils is less unexpected. Fixed to return a slice with possible nil entries (and mentioned that in the comments above).

module_ctx.go
```diff
@@ -775,17 +775,17 @@ type TopDownMutatorContext interface {
...
+	// dependency (some entries may be nil).  Does not affect the ordering of the current mutator
+	// pass, but will be ordered correctly for all future mutator passes.
+	AddDependency(module Module, tag DependencyTag, name ...string) []Module
+
+	// AddReverseDependency adds a dependency from the destination to the given module.  It returns
+	// a module for reverse dependency.  Does not affect the ordering of the current mutator pass,
+	// but will be ordered correctly for all future mutator passes.  All reverse dependencies for
+	// a destination module are collected until the end of the mutator pass, sorted by name, and
+	// then appended to the destination module's dependency list.
+	AddReverseDependency(module Module, tag DependencyTag, name string) Module
```
colincross, line 788: This one shouldn't return the module, since modules shouldn't ever be accessing the modules that depend on them.

module_ctx.go
```diff
@@ -818,25 +818,27 @@ type BottomUpMutatorContext interface {
...
-	// ordered in the same orderas they were listed in CreateVariations, and AddInterVariantDependency does not change
-	// that ordering, but it associates a DependencyTag with the dependency and makes it visible to VisitDirectDeps,
-	// WalkDeps, etc.
-	AddInterVariantDependency(tag DependencyTag, from, to Module)
+	// AddInterVariantDependency adds a dependency between two variants of the same module.
+	// It returns a module for the other variant.  Variants are always ordered in the same order
+	// as they were listed in CreateVariations, and AddInterVariantDependency does not change that
+	// ordering, but it associates a DependencyTag with the dependency and makes it visible to
+	// VisitDirectDeps, WalkDeps, etc.
+	AddInterVariantDependency(tag DependencyTag, from, to Module) Module
```
colincross, line 841: Returning Module here is redundant, since it was already passed in, but doesn't hurt if you think it helps consistency.

skvadrik:
Thanks! Please submit your change and I will rebase.

module_ctx.go
skvadrik: Reverted changes related to `AddReverseDependency`.

module_ctx.go
skvadrik: No, on a closer look I think it's useless. Reverted changes for AddInterVariantDependency.

module_ctx.go
```diff
@@ -775,10 +775,10 @@ type TopDownMutatorContext interface {
...
 	BaseMutatorContext
 
-	// AddDependency adds a dependency to the given module.
-	// Does not affect the ordering of the current mutator pass, but will be ordered
-	// correctly for all future mutator passes.
-	AddDependency(module Module, tag DependencyTag, name ...string)
+	// AddDependency adds a dependency to the given module.  It returns a slice of modules for each
+	// dependency (some entries may be nil).  Does not affect the ordering of the current mutator
+	// pass, but will be ordered correctly for all future mutator passes.
+	AddDependency(module Module, tag DependencyTag, name ...string) []Module
```
colincross, line 781: Can you add to each of these:

```
	//
	// If the mutator is parallel (see MutatorHandle.Parallel), this method will pause until the
	// new dependencies have had the current mutator called on them.  If the mutator is not
	// parallel this method does not affect the ordering of the current mutator pass, but will
	// be ordered correctly for all future mutator passes.
```

skvadrik: Done!

colincross (approved):

PR #319, opened 2020-08-31
Allow cloned modules to propagate arbitrary data through immutable pointers

colincross:
Support propagating arbitrary data between cloned modules by blindly
copying pointers tagged with `blueprint:"immutable_ptr"`.  The type
of the property must be a pointer, and the value referenced to must be
nil or never changed, as the value will be shared by multiple modules.
Unfortunately Go has no way to enforce immutability on a type, so this
has to be on the honor system.

Test: TestCloneProperties, TestCloneEmptyProperties, TestZeroProperties, TestExtendProperties
Change-Id: I488de0005c7883e5180b7ffeb2d69e4d1f9aea44

### DIFF
--- a/proptools/extend.go
+++ b/proptools/extend.go
@@ -273,13 +273,15 @@ func extendPropertiesRecursive(dstValues []reflect.Value, srcValue reflect.Value
 			}
 		}
 
-		// Step into source pointers to structs
-		if isStructPtr(srcFieldValue.Type()) {
-			if srcFieldValue.IsNil() {
-				continue
-			}
+		if !HasTag(srcField, "blueprint", "immutable_ptr") {
+			// Step into source pointers to structs
+			if isStructPtr(srcFieldValue.Type()) {
+				if srcFieldValue.IsNil() {
+					continue
+				}
 
-			srcFieldValue = srcFieldValue.Elem()
+				srcFieldValue = srcFieldValue.Elem()
+			}
 		}
 
 		found := false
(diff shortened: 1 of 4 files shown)


proptools/extend_test.go
```diff
@@ -1451,9 +1506,11 @@ func TestExtendMatchingProperties(t *testing.T) {
 func check(t *testing.T, testType, testString string,
 	got interface{}, err error,
 	expected interface{}, expectedErr error) {
+	t.Helper()
 
 	printedTestCase := false
 	e := func(s string, expected, got interface{}) {
+		t.Helper()
```
lizkammer, line 1513: This function is not being passed t, does it still need the t.Helper designation?

proptools/extend.go
```diff
@@ -309,6 +311,23 @@ func extendPropertiesRecursive(dstValues []reflect.Value, srcValue reflect.Value
...
+			if HasTag(srcField, "blueprint", "immutable_ptr") {
+				if !HasTag(dstField, "blueprint", "immutable_ptr") {
+					return extendPropertyErrorf(propertyName, `mismatched blueprint:"immutable_ptr" tag`)
+				}
+				if srcFieldValue.Type() != dstFieldValue.Type() {
+					return extendPropertyErrorf(propertyName, "mismatched types %s and %s",
+						dstFieldValue.Type(), srcFieldValue.Type())
+				}
+				if !srcFieldValue.IsZero() && !dstFieldValue.IsZero() {
+					return extendPropertyErrorf(propertyName, `overwrite immutable pointer`)
```
lizkammer, line 323: Optional: 

```suggestion
					return extendPropertyErrorf(propertyName, `cannot overwrite immutable pointer`)
```

proptools/extend_test.go
colincross: Yes, any function that is called multiple times and contains a t.Error method should use t.Helper() so that the test failure points to the line that called the helper function.

colincross:
I'm going to close this one, I think anything this could be used for would be better handled with Providers from #309.

PR #320, opened 2020-09-09
Enable variants of bootstrap modules

colincross:
The primary builder may want to create variants of bootstrap modules if they need to fit in to the primary builder's dependency graph.  Enable arbitrary variants of bootstrap modules by only running the module's actions on the primary variant and then copying the result to any other variants that exist.


colincross:
This enables the fix for the linux_bionic variant issue that got r.android.com/1410908 reverted.

danw (approved):

PR #326, opened 2020-10-16
Identify the type of the list for bpdocs

lizkammer:
Test: go test bpdoc tests
Test: m soong_docs
Change-Id: I6a4a916e1f72b3fc702da90c32a2eddca70b3bac

### DIFF
--- a/bootstrap/bpdoc/bpdoc_test.go
+++ b/bootstrap/bpdoc/bpdoc_test.go
@@ -6,6 +6,11 @@ import (
 	"testing"
 )
 
+type simpleProp struct {
+	name string
+	typ  string
+}
+
 type parentProps struct {
 	A string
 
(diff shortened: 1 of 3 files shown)


bootstrap/bpdoc/bpdoc_test.go
```diff
@@ -6,6 +6,11 @@ import (
 	"testing"
 )
 
+type simpleProp struct {
```
gyias, line 9: simpleProp sounds like this is one of the structs being tested. What about something like propInfo?

gyias (commented):
LGTM, just one nit.

bootstrap/bpdoc/bpdoc_test.go
lizkammer: Good idea!

gyias (approved):

PR #327, opened 2020-10-27
Improve formatting of some error messages.

marstj:
In particular the formatting of dependencies with variants which lacked
braces that caused them to float together with the dependency names.

Also add some context to the ReplaceDependenciesIf panic message.

Test: m

### DIFF
--- a/context.go
+++ b/context.go
@@ -1420,7 +1420,7 @@ func (c *Context) prettyPrintVariant(variations variationMap) string {
 		}
 	}
 
-	return strings.Join(names, ", ")
+	return strings.Join(names, ",")
 }
 
 func (c *Context) prettyPrintGroupVariants(group *moduleGroup) string {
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -1420,7 +1420,7 @@ func (c *Context) prettyPrintVariant(variations variationMap) string {
 		}
 	}
 
-	return strings.Join(names, ", ")
+	return strings.Join(names, ",")
```
lizkammer, line 1423: With the space removed, it looks like the line wrapping becomes a bit awkward -- why are we removing it?

marstj: The intention is to have a tighter grouping of variations in a variant than between variants in a list. Iow I think it makes it easier to read

libperfetto_hprof{os:android,image:,arch:arm64_armv8-a,link:shared,version:,art:debug}, libart-disassembler{os:android,image:,arch:arm64_armv8-a,link:shared,version:,art:debug}, libadbconnection{os:android,image:,arch:arm64_armv8-a,link:shared,version:,art:debug}, libart{os:android,image:,arch:arm64_armv8-a,link:shared,version:,art:debug}

than

libperfetto_hprof{os:android, image:, arch:arm64_armv8-a, link:shared, version:, art:debug}, libart-disassembler{os:android, image:, arch:arm64_armv8-a, link:shared, version:, art:debug}, libadbconnection{os:android, image:, arch:arm64_armv8-a, link:shared, version:, art:debug}, libart{os:android, image:, arch:arm64_armv8-a, link:shared, version:, art:debug}

marstj: So imo this more restrained use of spaces actually improves readability overall, but if you prefer spaces here or in other places just let me know and I'll add them.

lizkammer (approved):

context.go
marstj: Thank you. Shall I now submit it myself? I don't find a button for that anywhere.

PR #331, opened 2020-11-11
Export Glob method from Context.

lizkammer:
This allows use of globs outside of the context of a Singleton or Module.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1155,7 +1155,7 @@ func (c *Context) findBuildBlueprints(dir string, build []string,
 		var matches []string
 		var err error
 
-		matches, err = c.glob(pattern, nil)
+		matches, err = c.Glob(pattern, nil)
 
 		if err != nil {
 			errs = append(errs, &BlueprintError{
@@ -1197,7 +1197,7 @@ func (c *Context) findSubdirBlueprints(dir string, subdirs []string, subdirsPos 
 		var matches []string
 		var err error
 
-		matches, err = c.glob(pattern, nil)
+		matches, err = c.Glob(pattern, nil)
 
 		if err != nil {
 			errs = append(errs, &BlueprintError{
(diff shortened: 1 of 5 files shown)


colincross:
What's the use case for this?

You would have to be careful using ctx.Glob directly, if you use it after the generateGlobNinjaFile call in command.go you'll be missing dependencies.  You could make it an error to call ctx.Glob after ctx.Globs had already returned the final list of globs.

lizkammer:
> What's the use case for this?

Exploring this as an option for queryview in soong -- also looking into refactoring queryview such that this would be unnecessary.

> 
> You would have to be careful using ctx.Glob directly, if you use it after the generateGlobNinjaFile call in command.go you'll be missing dependencies. You could make it an error to call ctx.Glob after ctx.Globs had already returned the final list of globs.

Can you give more detail about what dependencies would be missing? It looks like generateGlobNinjaFile is writing out the cached globs?

jin:
(accidentally marked as ready, my bad!)

lberki:
In general, does anyone else other than Colin know how globs are handled?

(I added a question to the doc fixit spreadsheet)

colincross:
> Can you give more detail about what dependencies would be missing? It looks like generateGlobNinjaFile is writing out the cached globs?

`generateGlobNinjaFile` writes rules to `out/soong/.bootstrap/build-globs.ninja` that rerun each glob, writing lists of matching files to outputs like `out/soong/.glob/art/adbconnection/LICENCE.glob`.  `globSingleton` calls `ctx.AddNinjaFileDeps` on each of those output files.  This causes the rule that runs `soong_build` to generate `out/soong/build.ninja` depend on each of those glob list files.  If you add a new glob after `globSingleton` has run or after `generateGlobNinjaFile` has been called, there will be no dependency from `out/soong/build.ninja` to the file list, and `soong_build` won't rerun if a matching file is added or deleted.

lberki:
Closing this since we won't be merging this as the project is being archived.

PR #332, opened 2020-11-20
bpmodify: Do not sort automatically

jooyunghan:
bpmodify automatically sorts touched list when it is originally sorted.
Even when a user doesn't want this behavior there's no way to avoid
sorting. For example, lists with version numbers.

Now bpmodify sorts the list only if `-s` is given.

Test: bpmodify_test.go
Change-Id: If2fe9bc871a463a47dd3c0b52982b34c9fde05f0

### DIFF
--- a/bpmodify/bpmodify_test.go
+++ b/bpmodify/bpmodify_test.go
@@ -206,6 +206,48 @@ var testCases = []struct {
 		"",
 		"dep3,dep4",
 	},
+	{
+		`
+		cc_foo {
+			name: "foo",
+			versions: ["1", "2"],
+		}
+		`,
+		`
+		cc_foo {
+			name: "foo",
+			versions: [
+				"1",
+				"2",
+				"10",
+			],
+		}
+		`,
+		"versions",
+		"10",
+		"",
+	},
+	{
+		`
+		cc_foo {
+			name: "foo",
+			deps: ["bar-v1-bar", "bar-v2-bar"],
+		}
+		`,
+		`
+		cc_foo {
+			name: "foo",
+			deps: [
+				"bar-v1-bar",
+				"bar-v2-bar",
+				"bar-v10-bar",
+			],
+		}
+		`,
+		"deps",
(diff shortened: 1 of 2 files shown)


colincross:
I updated the build actions in #333 but this seems to rerun the actions as they were defined in your commit, not after the merge.  Can you rebase your change?

bpmodify/bpmodify.go
```diff
@@ -240,7 +238,7 @@ func processParameter(value parser.Expression, paramName, moduleName string,
 		modified = modified || m
 	}
 
-	if (wasSorted || *sortLists) && modified {
```
colincross, line 243: I hadn't realized this defaults to only operating on already-sorted lists.  I don't understand why this is in issue for you.  Why is a list with version numbers different?  Would switching to numerical sort solve  your problem?

jooyunghan: >  I don't understand why this is in issue for you. Why is a list with version numbers different? 
adding "10" to `versions: ["1", "2"]` resulting `versions: ["1", "10", "2"]`.

> Would switching to numerical sort solve your problem?
Yes, that's my original trial. (Adding `-n` to do numerical sort). But I found that passing `-s` explicitly to sort is better. Do you want to keep the original behavior (sorting already-sorted list)?

jooyunghan:
Added -n option to sort numerically.

btw, I'm thinking this should be automatic, just like "wasSorted". We can examine if the list is numeric before sorting or checking sortedness. But there's a corner case like the list which is alphabetically sorted but not numerically.

jooyunghan:
Instead of adding a new option (`-n`), I changed the default sorting behavior to support numerical sort if all elements are numbers(technically, it can be converted using `strconv.Atoi`).

jooyunghan:
Rebased and changed the behavior to sort numerically. Can you please take a look?

jooyunghan:
@colincross please take a look.

colincross (approved):

PR #333, opened 2020-11-24
Fix github builds for deprecated set-env and add-path commands

colincross:
actions/setup-go@v2.0.3 seems to cause issues with
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Change the precise versions to major versions to get updates.

Also fix the "Install ninja" action to use $GITHUB_PATH.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,13 +17,13 @@ jobs:
     steps:
 
     - name: Set up Go ${{ matrix.go }}
-      uses: actions/setup-go@v2.0.3
+      uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go }}
       id: go
 
     - name: Check out code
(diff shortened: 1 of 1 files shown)


danw (approved):
https://opensource.google/docs/github/services/#actions (go/github-actions internally) says that we need to use tagged releases instead of branches, but I guess that `v2` is actually a tag that gets updated? 🤷‍♂️

colincross:
> https://opensource.google/docs/github/services/#actions (go/github-actions internally) says that we need to use tagged releases instead of branches, but I guess that `v2` is actually a tag that gets updated, rather than a branch?

That mentions "Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.".  Having v2 as a tag that changes is a little odd though.

PR #334, opened 2020-11-30
Invalidate module group cache if deps are modified

lizkammer:
PreSingletons run after the blueprints have been parsed and can run         
VisitAllModules; however, this seeds the cache of sorted modules which   
may change after mutators have run. This causes recomputation of the        
cache if any deps have been modified since the last time it was run.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1549,6 +1549,9 @@ func (c *Context) resolveDependencies(ctx context.Context, config interface{}) (
 		if len(errs) > 0 {
 			return
 		}
+		// if any presingletons visited all modules, the cache would be populated by only modules that exist pre-mutators.
+		// we clear the cache so that future calls to visit all modules will be accurate
+		c.clearSortedModuleGroupsCache()
 
 		errs = c.updateDependencies()
 		if len(errs) > 0 {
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -1549,6 +1549,9 @@ func (c *Context) resolveDependencies(ctx context.Context, config interface{}) (
 		if len(errs) > 0 {
 			return
 		}
+		// if any presingletons visited all modules, the cache would be populated by only modules that exist pre-mutators.
+		// we clear the cache so that future calls to visit all modules will be accurate
+		c.clearSortedModuleGroupsCache()
```
colincross, line 1554: Whenever a new module is added we increment the atomic c.depsModified variable, can we make sortedModuleGroups() automatically clear the cache if that variable is set, or add a new one if the interactions with the existing users of that variable are a problem?

lizkammer: Good idea. Done.

lizkammer: Further updated to address your comments about depsModified being reset.

colincross (approved):

jin:
LGTM

PR #336, opened 2021-01-12
Introduce post-mutator singletons.

jin:
Post-mutator singletons, as the name describes, runs immediately after registered mutators finish. With PreSingletons, this enables singletons to run directly before and after mutators, all within ResolveDependencies, and not interfering with the GenerateBuildActions phase.

Used for https://r.android.com/1527838


jin:
Not needed anymore.

PR #339, opened 2021-01-20
Remove redundant stat/lstat calls from glob

colincross:
Glob was calling IsSymlink and IsDir on each visited directory entry,
which resulted in an lstat and then a stat call on each.

Instead, use lstat when not following symlinks and use stat when
following symlinks, then use the result to check if the entry is a
directory.

Test: glob_test.go
Change-Id: I83d769e2de64ce8221e952e5204d365aeaf47687

### DIFF
--- a/pathtools/glob.go
+++ b/pathtools/glob.go
@@ -76,24 +76,18 @@ func startGlob(fs FileSystem, pattern string, excludes []string,
 	}
 
 	for i, match := range matches {
-		isSymlink, err := fs.IsSymlink(match)
+		var info os.FileInfo
+		if follow == DontFollowSymlinks {
+			info, err = fs.Lstat(match)
+		} else {
+			info, err = fs.Stat(match)
+		}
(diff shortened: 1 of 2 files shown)


pathtools/glob.go
```diff
@@ -76,24 +76,18 @@ func startGlob(fs FileSystem, pattern string, excludes []string,
...
-				}
-			}
-			if err != nil {
-				return nil, nil, fmt.Errorf("%s: %s", match, err.Error())
-			}
 
-			if isDir {
-				matches[i] = match + "/"
-			}
+		if info.Mode()&os.ModeDir != 0 {
```
lizkammer, line 89: any reason to not use the info.IsDir() abbreviation throughout?

lizkammer (approved):

pathtools/glob.go
colincross: Whoops, missed this, will fix.

PR #345, opened 2021-02-15
Fix ReplaceExtension

tweksteen:
ReplaceExtension had an unexpected behaviour when the file did not have
an extension. In certain cases, the final path would be severely
trimmed: out/.intermediates/my_file would become out/.new_extension.
Explicitly handle the case by appending the new extension.

Test: Run checkbuild on Android Soong
Change-Id: Ie27a98845894cfaee5af5e2a02d44168c40ed821

### DIFF
--- a/Blueprints
+++ b/Blueprints
@@ -70,6 +70,7 @@ bootstrap_go_package {
     testSrcs: [
         "pathtools/fs_test.go",
         "pathtools/glob_test.go",
+        "pathtools/lists_test.go",
     ],
 }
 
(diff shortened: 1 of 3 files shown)


colincross (commented):
Sorry, we're in progress with moving the source of truth for blueprint to AOSP, can you upload your change there?

colincross:
Nevermind, I used copybara to import it as https://android-review.googlesource.com/c/platform/build/blueprint/+/1591572

tweksteen:
Great, thanks.

PR #346, opened 2021-03-08
Add ShellArgEscape(string)

jooyunghan:
ShellEscape(string) doesn't escape a string with spaces like "arg1
arg2". However, when we want to escape a string to use it as an
argument, then strings with spaces should be escaped.

For example,

  "command " + ShellArgEscape("a b")

becomes "command 'a b'" so that "command" will get "a b" as a single
argument.

Bug: 182092664
Test: Added tests to escape_test.go
Change-Id: I8f88c18bc4f9f7aacfe9e701b8f0876dd8b9a8c3

### DIFF
--- a/proptools/escape.go
+++ b/proptools/escape.go
@@ -56,7 +56,7 @@ func ShellEscapeList(slice []string) []string {
 
 }
 
-// ShellEscapeList takes string that may contain characters that are meaningful to bash and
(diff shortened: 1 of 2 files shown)


jin:
cc @lberki 

Blueprint's source of truth has moved to AOSP main. Please submit your change there instead.

jooyunghan:
@jin Thank you!

